PR #106, opened 2023-08-24
Dhcp issues

jhughesbiot:
Fixes issue: https://github.com/google/test-run/issues/105

The increased new lease wait time and configurable option will help mitigate issue: https://github.com/google/test-run/issues/104

### DIFF
--- a/modules/test/conn/conf/module_config.json
+++ b/modules/test/conn/conf/module_config.json
@@ -49,6 +49,7 @@
         "expected_behavior": "The device under test accepts IP addresses within all ranges specified in RFC 1918 and communicates using these addresses.  The Internet Assigned Numbers Authority (IANA) has reserved the following three blocks of the IP address space for private internets.  10.0.0.0 - 10.255.255.255.255 (10/8 prefix). 172.16.0.0 - 172.31.255.255 (172.16/12 prefix).  192.168.0.0 - 192.168.255.255 (192.168/16 prefix)",
         "required_result": "Required",
         "config": {
+          "lease_wait_time_sec": 60,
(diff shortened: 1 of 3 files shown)


jhughesbiot:
@jboddey 
> I have had another thought about the changes in this PR. We should not have to wait beyond the time of the lease for the device to obtain an IP address. The device should attempt to renew the lease at T1 (50% of the time through the current lease). For tests where the subnet changes, the device should attempt to find another DHCP server at T2 (87.5% of the time through the current lease). So if the device has not got an IP address by the end of the current lease, I would consider this a non-compliance.

Where did these percentages come from?

I think we should track this change via a separate issue and merge this in so we can get the current issues resolved and make the test behavior change to it's own PR.

jboddey:
> @jboddey 
> 
> > I have had another thought about the changes in this PR. We should not have to wait beyond the time of the lease for the device to obtain an IP address. The device should attempt to renew the lease at T1 (50% of the time through the current lease). For tests where the subnet changes, the device should attempt to find another DHCP server at T2 (87.5% of the time through the current lease). So if the device has not got an IP address by the end of the current lease, I would consider this a non-compliance.
> 
> 
> 
> Where did these percentages come from?
> 
> 
> 
> I think we should track this change via a separate issue and merge this in so we can get the current issues resolved and make the test behavior change to it's own PR.
> 
> 
> 
> 
> 
> 
> 
> 

The times are from the DHCP RFC. Are you experiencing this issue on one of your devices?

jhughesbiot:
I haven't recreated these issues on devices I have tested.

PR #127, opened 2023-09-26
Test fixes

jhughesbiot:
Fixes the following issues:

TLS Module:
 - TLS client details reported results from the monitor when only startup had valid packets

Connection Module:
 - Different devices reported DHCP Request options in different orders causing inconsistent results.

### DIFF
--- a/modules/test/tls/python/src/tls_module.py
+++ b/modules/test/tls/python/src/tls_module.py
@@ -97,7 +97,7 @@ class TLSModule(TestModule):
     elif monitor_result[0] and startup_result[0] is None:
       result = True, monitor_result[1]
     elif startup_result[0] and monitor_result[0] is None:
-      result = True, monitor_result[1]
+      result = True, startup_result[1]
     else:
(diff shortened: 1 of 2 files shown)


modules/test/tls/python/src/tls_module.py
```diff
@@ -97,7 +97,7 @@ def _validate_tls_client(self, client_ip, tls_version):
     elif monitor_result[0] and startup_result[0] is None:
       result = True, monitor_result[1]
     elif startup_result[0] and monitor_result[0] is None:
-      result = True, monitor_result[1]
+      result = True, startup_result[1]
```
jboddey, line 100: Is this specifying the packet capture to check in? Could we check both to be safe?

jhughesbiot: No, it checks both, this just corrects the details of the test.  If client packets were detected in the startup capture but not in the monitor, it was giving a compliant test but the details said no packets were detected because that was the results of the monitor capture.

PR #130, opened 2023-10-03
Release: 1.0.2

jboddey:
Makes the following changes:
 - #140 
 - #144 
 - #147 
 - #155 
 - Cancelled state appears if stop button is pressed
 - Remove outdated device configuration guide
 - Adds device DHCP pre-req to README
 - Details how to configure a device to achieve a compliant result
 - Inform users that device should be powered off before starting testrun
 - Adds uninstall instructions and uninstall scripts
 - Remove use of runtime, run infinitely until user stops Testrun
 - Removes duplicate docs in connection test module
 - Fix PDF report url


jhughesbiot:
Logging for modules appears to be broken and/or inconsistent.  Main log during runtime shows logs that do not appear in the module log and vice versa.  Connection module is probably the best example of this where DHCP lease information shows up inconsistently between the two logs.

PR #131, opened 2023-10-10
Ipv6 updates

jhughesbiot:
Improves the IPv6 testing with the following updates:

 - Increase advertise frequency
 - Decrease lifetime
 - Scan DHCP capture instead of monitor capture
 - Disable RADVD server in dhcp2 module


modules/network/dhcp-1/python/src/grpc_server/radvd_server.py
```diff
@@ -22,29 +22,42 @@
 class RADVDServer:
   """Represents the RADVD Server"""
 
-  def __init__(self):
+  def __init__(self, enabled=True):
     global LOGGER
-    LOGGER = logger.get_logger(LOG_NAME, 'dhcp-1')
+    LOGGER = logger.get_logger(LOG_NAME, 'dhcp-2')
```
jboddey, line 27: @jhughesbiot Is this supposed to be dhcp-2 in the dhcp-1 module?

jhughesbiot: Yes, this has been updated.

jboddey (approved):

Issue #104, opened 2023-08-24
Device fails DHCP tests if it obtains IP from secondary DHCP server initially


jboddey:
**Describe the bug**
A device has received it's first IP from the secondary DHCP server. When the IP changes tests run, the secondary DHCP server is taken down. The device is then unable to reach the secondary DHCP server for some time. It follows the process to then identify a new DHCP server but this is not quick enough and instead fails the private address space test.

**Expected behavior**
Perhaps the DHCP secondary server should be disabled by default rather than being disabled during the connection test module. The dhcp_failover test can enable the secondary server

**Error logs**
[If applicable, provide a log from https://gist.github.com/](https://gist.github.com/jboddey/bfdf86c7f688c2a81d6f1eee06020120)

**Environment (please provide the following information about your setup):**
 - OS: Kali linux
 - Version unknown
 - Additional hardware: 1x USB device adapter and 1x ethernet

PR #167, opened 2023-10-26
Conn test fixes

jhughesbiot:
Adds improvements for dhcp related tests, specifically around secondary DHCP server picking up the original lease, issue https://github.com/google/testrun/issues/104

Adds configuration parameters around the lease wait time which supersedes PR
https://github.com/google/testrun/pull/106


jboddey:
I'm happy with this technically. Still unsure of the reasoning behind waiting for the lease. All devices I have tested are able to renew their lease before the lease expires. Are we loosening the test requirements for specific devices here? Could you let us know which devices you have issues with?

jhughesbiot:
Maybe you've just gotten lucky?  All devices I have tested require the full lease expiration time to complete plus some level of buffer to get a new lease, especially with subnet change. I don't think changing the timing of the lease does anything to modify the test requirements as the test is only that they can achieve these IP addresses and says nothing about specific timing requirements adherent to any RFC beyond accepting the IP address within those ranges.  

Some devices I have tested worked just fine and got a new lease within the lease time (30 sec) of expiration but others need significantly longer.  This could be device specific for any number of reasons so having a configurable timer starting at 1 minute to add more stability and less false failures is a necessity.

Because this is just an added timer, it wouldn't cause any changes to devices that accept leases faster, just gives more time to devices that need a little longer to accept/change their DHCP lease information which does not conflict with the current test plan requirements.

jboddey (approved):

PR #169, opened 2023-10-27
Resolve timezone issue in connection modules

jboddey:
Now that the timezone has been set to the local timezone within test and network containers, the test code within the connection module does not function correctly - it is expecting the local time of the container to be UTC (the same as the time zone used in the DHCP server lease file). Therefore, wait times etc are not correct. This pull request manually converts local time back to UTC time when making date time comparisons with the DHCP lease expiration times.

### DIFF
--- a/framework/python/src/common/device.py
+++ b/framework/python/src/common/device.py
@@ -14,8 +14,9 @@
 
 """Track device object information."""
 
-from typing import Dict
+from typing import Dict, List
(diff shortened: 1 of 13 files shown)


jboddey:
This pull request also fixes the following bugs:
 - Disables the baseline module by default
 - Adds in default test descriptions if none have been specified

noursaidi (approved):

PR #182, opened 2023-11-13
Merge dev into main (Release 1.1-alpha)

jboddey:


jhughesbiot:
Bad merge conflict resolution here removed remote UI access https://github.com/google/testrun/commit/bc1830ada02a313f6df597c7f4a3636ea43c5525#diff-efd6ae663231fce881ad51b3f743972b3929c93c52d1e8b8d48d9faee79d8e73

jboddey:
> Bad merge conflict resolution here removed remote UI access [bc1830a#diff-efd6ae663231fce881ad51b3f743972b3929c93c52d1e8b8d48d9faee79d8e73](https://github.com/google/testrun/commit/bc1830ada02a313f6df597c7f4a3636ea43c5525#diff-efd6ae663231fce881ad51b3f743972b3929c93c52d1e8b8d48d9faee79d8e73)

I'd like for us to overwrite this for now and initiate the remote access as a new feature branch with assistance from the UI team.

jhughesbiot:
Why would we want to remove a feature already present now?

jboddey:
> Why would we want to remove a feature already present now?

The UI code is currently part of another repository and each update requires copying of the code from that one to here. It wouldn't be ideal to have to manage code differences between the two repositories. I will request that the UI team implement the code change and then merge their repository into ours so this shouldn't occur in the future

jhughesbiot:
I guess if we have to get the UI updated in a different repo first, then let's do that.

jhughesbiot (approved):

modules/test/tls/conf/module_config.json
jhughesbiot, line 1: Why are we removing TLS 1.3 tests? TLS 1.3 server is a part of the official test plan.  I can see removing TLS 1.3 client potentially but I don't see any reason for the removal of these tests.

jboddey: Disabling in the module config did not seem to work so I removed them from the module config for now. We have been given the direction to remove recommended tests from Testrun. These are being removed from the test plan as of v4.5.

PR #203, opened 2023-12-08
Add device configuration to PDF report

jboddey:
Before: 

image

After:

image


jboddey:
Ignore the 'Unknown' module. This is an old configuration that has the 'DHCP' module enabled, which is pending removal.

jhughesbiot (approved):

PR #205, opened 2023-12-08
Fix remote ui

jhughesbiot:
Add missing commit from previous remote ui PR


jboddey (approved):

jhughesbiot:
Updated to fix report download link for remote connections

PR #207, opened 2023-12-13
Add UI and update README

jboddey:


README.md
jhughesbiot, line 1: Maybe change Device heading under minimum requirements to Device Under Test (DUT) to keep the naming conventions consistent

modules/test/tls/python/src/tls_util.py
```diff
@@ -380,8 +380,9 @@ def validate_tls_client(self, client_ip, tls_version, capture_file):
           tls_client_details += 'Completed handshake detected from server: '
           tls_client_details += result + '\n'
     else:
-      LOGGER.info('No client hello packets detected. Skipping')
-      tls_client_details = 'No client hello packets detected. Skipping'
+      LOGGER.info('No client hello packets detected')
+      tls_client_details = 'No client hello packets detected'
+      return True, tls_client_details
```
jhughesbiot, line 385: Returning true here would make this test pass even though the device didn't make any client connections.  This needs to remain in a skipped state by returning None

jboddey: This is the intention for now until we finalize the decision around 'Feature not present' or Skipped etc.

jhughesbiot: Understood but changing a test to pass when it hasn't cannot be the correct decision.  This will cause invalid and unusable results in the report.

jboddey: I have reverted this change and will look at this again in Q1.

README.md
jboddey: Changed

jhughesbiot (approved):

Issue #221, opened 2024-01-08
Add limit to connection module wait time


jboddey:
**Describe the bug**
The connection module is suggesting that device lease times are longer than expected. This makes the system hang for a long period of time and then proceeds to effect the results of the connection module.

**To Reproduce**
1. Start Testrun
2. Run connection module

**Expected behavior**
Connection module should wait up to the max lease time setting instead of the reported lease time from the DHCP server.

**Error logs**
https://gist.github.com/jboddey/28167d81cb99c27018197cfce2cb661e

jboddey:
Merged. Due to release in v1.1.1 at end of January.

PR #223, opened 2024-01-10
Bug/221 conn module lease times

jhughesbiot:
Resolves Issue: https://github.com/google/testrun/issues/221
- Convert DHCP lease times to local time zone before expiration time calculated
- Add max wait time when waiting for lease expiration

### DIFF
--- a/modules/test/conn/python/requirements.txt
+++ b/modules/test/conn/python/requirements.txt
@@ -1,2 +1,3 @@
 pyOpenSSL
 scapy
+python-dateutil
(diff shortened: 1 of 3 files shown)


jboddey (approved):

jboddey:
@jhughesbiot We want to merge this into v1.1.1 branch but having changing the base, it seems to have brought across all of the dev commits so far. Could you see what the issue might be?

jhughesbiot:
Because it's branched from dev originally as usual process.  If we want to patch just this release I'd have to branch off of that first and then manually make these changes again.

PR #231, opened 2024-01-19
Module reports

jhughesbiot:
- Adds DNS module markdown report
- Adds framework test modules to generate markdown reports
- Resolves markdown reports from test modules and adds them to final report


jboddey:
Hi Jonathan, I am currently getting this error when running this branch:

`Traceback (most recent call last):
  File "/testrun/python/src/run.py", line 77, in <module>
    run()
  File "/testrun/python/src/run.py", line 73, in run
    DNSModuleRunner(args.module.strip())
  File "/testrun/python/src/run.py", line 42, in __init__
    self._test_module.generate_module_report()
  File "/testrun/python/src/dns_module.py", line 71, in generate_module_report
    max_data_length = max(len(row['Data']) for row in dns_table_data)
ValueError: max() arg is an empty sequence
`

jhughesbiot:
Confirmed via unit testing this is a no DNS scenario.  Unit test added around this scenario and should be corrected.

PR #239, opened 2024-01-24
Nmap report

jhughesbiot:
- Adds the NMAP module report
- Updates to the report formatting for module reports
- Adds unit testing around html report creation
- Updates and normalizes unit testing structure for existing test modules
- Adds gateway capture to TLS client test.


jboddey:
@jhughesbiot could we remove closed ports from the table?

jhughesbiot:
I can but I was under the impression we wanted to show the full results of the scan.  If this isn't the case, I can filter out for only open ports.  However, if we have an empty table with no details because nothing is open, it might look broken.

jboddey:
> I can but I was under the impression we wanted to show the full results of the scan. If this isn't the case, I can filter out for only open ports. However, if we have an empty table with no details because nothing is open, it might look broken.

Could we perhaps remove the table and provide test stating 'No ports have been detected open' if none are found? Just with all ports open, the nmap part of the report takes up more than a page for me.

jhughesbiot:
Yeah, that sounds good, I'll check for no open ports and update accordingly.  I wouldn't be too concerned with going over a page though, the DNS equivalent table can cover multiple pages.

jhughesbiot:
Updated table with closed ports removed:

image

Updated report if no ports are detected open:
image

jboddey (approved):

PR #256, opened 2024-02-05
Merge release v1.1.2 into main

jboddey:


jboddey:
test_api will be fixed when PR #262 is merged into release/v1.1.2

jhughesbiot:
Download button does not work with remote connections, still hard coded to localhost.  Should be updated to use similar methods implemented https://github.com/google/testrun/pull/256/commits/a233292a811f39c3f76e2a0a97fb9910d1c4ea42

image

jhughesbiot:
Version needs to be updated to match this release:

In cmd/package file:
```
# Rename the .deb file
mv make.deb testrun_1-1-1_amd64.deb
```

And framework/python/src/common/testreport.py

```
  def generate_page(self, json_data, page_num, max_page):
    # Placeholder until available in json report
    version = 'v1.1.1 (2024-01-31)'
    page = '<div class="page">'
    page += self.generate_header(json_data)
    if page_num == 1:
      page += self.generate_summary(json_data)
    page += self.generate_results(json_data, page_num)
    page += self.generate_footer(page_num,max_page,version)
    page += '</div>'
    if page_num < max_page:
      page += '<div style="break-after:page"></div>'
    return page
```

jhughesbiot:
> I have added api_url to system.json

This shouldn't need to be configured, it should auto detect just like the other download link.  This is likely a UI update to complete full remote connection feature.

jhughesbiot:
@jboddey I found an issue with the new container logging that was blocking the main run loop for the test container.  I moved this to it's own thread and pushed an update to correct.  Please confirm this is good on your end before merging.

jboddey:
> > I have added api_url to system.json
> 
> This shouldn't need to be configured, it should auto detect just like the other download link. This is likely a UI update to complete full remote connection feature.

If we want the remote access in this patch version then this implementation will have to do for now. Otherwise it will have to be delayed until v1.2

jboddey:
> > I have added api_url to system.json
> 
> This shouldn't need to be configured, it should auto detect just like the other download link. This is likely a UI update to complete full remote connection feature.

If we want the remote access in this patch version then this implementation will have to do for now. Otherwise it will have to be delayed until v1.2

jboddey:
> > I have added api_url to system.json
> 
> This shouldn't need to be configured, it should auto detect just like the other download link. This is likely a UI update to complete full remote connection feature.

If we want the remote access in this patch version then this implementation will have to do for now. Otherwise it will have to be delayed until v1.2

KrystsinaT (approved):

jhughesbiot (approved):

PR #257, opened 2024-02-06
Add steps to resolve to dev

jboddey:


framework/python/src/test_orc/test_case.py
```diff
@@ -24,3 +24,4 @@ class TestCase:  # pylint: disable=too-few-public-methods,too-many-instance-attr
   description: str = ""
   expected_behavior: str = ""
   required_result: str = "Recommended"
+  steps_to_resolve: list = field(default_factory=lambda: [])
```
OlgaMardvilko, line 27: Hi @jboddey 
Based on the API doc (https://docs.google.com/document/d/1XI-Jv-tEHRc9x9TOy9SBKixYSai6i9YmZYm0LFzkVv4/edit?resourcekey=0-LywLMRKxfE-5Ax9R0QbeaQ&tab=t.0#heading=h.fd0nu6hnb8q5) and provided mockoon data  we expect that the property name will be `recommendations`.
Here the FE PR where we already rely on this property: https://github.com/google/testrun/pull/267/files#diff-e762ffbb2dda3212534eb77279b33552b4bb10c2313c0a6cb7390b72904d6fe6R38

So, let discuss and decide where we change it: on BE or FE side.

OlgaMardvilko (changes requested):

framework/python/src/test_orc/test_case.py
jboddey: Noted, I will change this on the backend shortly

OlgaMardvilko (approved):

KrystsinaT (approved):
Reviewed.

PR #258, opened 2024-02-06
Add appropriate headings

jboddey:


jboddey:
Need to disable text overflow in test result descriptions before merging

jhughesbiot:
Changing the the headers to use the h1/2/3 will change the formatting of the module reports from markdown that use these and likely breaking the formatting there.  Unclear on the need to change the class vs updating the CSS for this header classes.  If this is necessary, the markdown translation formatting will need to be addressed before merging.

jboddey:
We need to change the class so that when it is converted into a PDF, the headers are suitable for screen readers. This is a blocking feature for our accessibility rating. We can deal with any issues in the module reports once release/v1.1.2 is merged into dev.

jhughesbiot:
If this change breaks other elements of the report, we shouldn't approve and merge until the break is addressed within the same PR.  No need to knowingly introduce bugs especially into a release branch.

jboddey:
Looking back at this, there is no plan to release the module reports in this release. Module reports will be released end of Q1 whilst we expect to release this end of this month at the latest. When the time comes, and we open the PR to merge release/v1.1.2 into dev (which may reveal these bugs) I will fix any CSS issues.

KrystsinaT (approved):

jboddey:
PDF report currently cuts off overflow of test descriptions. This should be updated to text-wrap or shorten test descriptions to improve accessibility.

PR #260, opened 2024-02-07
Feature/dev UI

sofyakurilova:
Merge dev_ui with dev after release


jboddey:
@sofyakurilova I have found a bug in this PR:
When the user saves the connection settings, the payload sent to the API is as follows:
`
{ "network": 
{
device_intf: "0", 
internet_intf: "1"
}}`

This results in invalid interfaces being saved in the system config. The UI should still return the actual interface name to the API.

jboddey:
The bug is now resolved. @KrystsinaT could you verify perhaps?

jboddey (approved):

PR #263, opened 2024-02-08
Refactoring

sofyakurilova:
- adds component store
- move dropdowns to separate component

8CrUpcPUKUiNmBR

### DIFF
--- a/modules/ui/src/app/pages/settings/general-settings.component.ts
+++ b/modules/ui/src/app/pages/settings/general-settings.component.ts
@@ -16,52 +16,43 @@
 import {
   Component,
   EventEmitter,
-  Input,
   OnDestroy,
   OnInit,
   Output,
 } from '@angular/core';
 import { FormBuilder, FormControl, FormGroup } from '@angular/forms';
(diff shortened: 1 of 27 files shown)


modules/ui/src/app/pages/settings/general-settings.component.ts
```diff
@@ -16,52 +16,43 @@
...
   public readonly CalloutType = CalloutType;
   public readonly EventType = EventType;
+  public readonly FormKey = FormKey;
   public settingForm!: FormGroup;
-  public isSubmitting = false;
-  public defaultInternetOption = {
-    key: '',
-    value: 'Not specified',
-  };
+  vm$ = this.settingsStore.vm$;
```
OlgaMardvilko, line 46: Could you please explain what is mean vm? Probably it is better to create name that will self explain.

OlgaMardvilko (approved):

modules/ui/src/app/pages/settings/general-settings.component.ts
sofyakurilova: Done

PR #270, opened 2024-02-13
Tls server bug

jhughesbiot:
Corrects bug when a TLS 1.3 server connection is detected but not a TLS 1.2, the tls 1.2 server test will result in a skip.


jboddey:
Hi Jonathan, I am assuming we also want this in the v1.1.2 release? If so, I have added the changes to that release/v1.1.2 branch

jhughesbiot:
Yes, there's no reason to not have it there as well.

PR #271, opened 2024-02-14
Enable save button when only one interface is available

sofyakurilova:
5r2orWvfrC8xpV7


OlgaMardvilko (commented):
I compare this PR with the original one that already marge in the dev branch and see some differences. Here the original PR for this bug: https://github.com/google/testrun/pull/252/files

Do we expect it?

OlgaMardvilko (approved):

PR #277, opened 2024-02-15
Re-enable protocol tests

jboddey:


jhughesbiot (approved):
Protocol tests are now run first if enabled to prevent any network changing from initial IP and port binding that could cause false failures with BACnet.

PR #280, opened 2024-02-19
Add arp inspection and dhcp snooping tests

jboddey:


jboddey:
@jhughesbiot Could you let me know what you think of these test additions and whether you agree on my interpretation of the test requirements from the test plan?

modules/test/conn/python/src/connection_module.py
```diff
@@ -74,6 +74,60 @@ def __init__(self, module):
...
+
+    if no_arp:
+      return None, 'No ARP packets from the device found'
+
+    return True, 'Device uses ARP'
+
+  def _connection_switch_dhcp_snooping(self):
+    LOGGER.info('Running connection.switch.dhcp_snooping')
+
+    disallowed_dhcp_types = [2, 4, 5, 6, 9, 10, 12, 13, 15, 17]
```
jhughesbiot, line 111: This doesn't appear like you are allowing any DHCP traffic from the device which does not match the test requirements.  Not sure what the intention here is for filtering this.

modules/test/conn/python/src/connection_module.py
```diff
@@ -160,6 +214,12 @@ def _connection_single_ip(self):
     else:
       return result, 'Device is using multiple IP addresses'
 
+  def _get_dhcp_type(self, packet):
+    for option in packet[DHCP].options:
+      # message-type, option 3 = DHCPREQUEST
```
jhughesbiot, line 219: Comment appears to be more aligned with the change on line 198 and doesn't fit here.

modules/test/conn/python/src/connection_module.py
jboddey: DHCP snooping is switch functionality to prevent a device from sending DHCP server messages on the network. We cannot add a managed switch with DHCP enabled so instead this test checks that the device does not send those DHCP server messages - pretending to be a switch. The array contains all IDs of the DHCP server messages.

modules/test/conn/python/src/connection_module.py
```diff
@@ -74,6 +74,60 @@ def __init__(self, module):
     # response = self.dhcp1_client.set_dhcp_range('10.10.10.20','10.10.10.30')
     # print("Set Range: " + str(response))
 
+  def _connection_switch_arp_inspection(self):
+    LOGGER.info('Running connection.switch.arp_inspection')
+
+    no_arp = True
+
+    # Read all the pcap files
+    packets = rdpcap(STARTUP_CAPTURE_FILE) + rdpcap(MONITOR_CAPTURE_FILE)
```
jhughesbiot, line 83: Inspecting the startup capture causes false failures because the two DHCP servers will get their own responses from the device with different IP's. 

image

Attempting to just inspect the monitor appears to show similar issues where the ARP who has packets from the dhcp servers during startup appear to propagate, maybe through arp caching into the monitor and still get responses from each IP address.  This may require a more complex process to detect bad ARP packets with the usage of DHCP failover because this does not appear to be a failure of the device.

modules/test/conn/python/src/connection_module.py
```diff
@@ -74,6 +74,60 @@ def __init__(self, module):
...
+        continue
+
+      # Get the ARP packet
+      arp_packet = packet[ARP]
+      no_arp = False
+
+      # Check MAC address matches IP address
+      if (arp_packet.hwsrc != self._device_mac or
+          arp_packet.psrc != self._device_ipv4_addr):
+        return False, 'Device is sending false ARP response'
```
jhughesbiot, line 101: Logic check needs to be updated so that we are checking if the expected IP is being used for the correct mac.  Currently this filters for any ARP that isn't the device mac or IP which isn't correct for arp insepection.

This fixes that and adds some useful logging for failures:
```
      # Check MAC address matches IP address
      if (arp_packet.hwsrc == self._device_mac and
          arp_packet.psrc != self._device_ipv4_addr):
        LOGGER.info(f'Bad ARP packet detected for MAC: {self._device_mac}')
        LOGGER.info(f'ARP packet IP {arp_packet.psrc} does not match {self._device_ipv4_addr}')
        return False, 'Device is sending false ARP response'
```

modules/test/conn/python/src/connection_module.py
jboddey: I'm not sure I'm following, I am not getting any failures when running this test on physical devices even with DHCP failover.

jhughesbiot: I can no longer recreate the failure I was seeing when this was reported.  I can't find any difference between my testing scenario so I'm ok moving forward with this.

PR #286, opened 2024-02-22
Improve dev process

jboddey:


cmd/install
```diff
@@ -16,7 +16,13 @@
 
 echo Installing application dependencies
 
-TESTRUN_DIR=/usr/local/testrun
+if [[ -z "${TESTRUN_DIR}" ]]; then
+  TESTRUN_DIR=/usr/local/testrun
+else
+  TESTRUN_DIR="${TESTRUN_DIR}"
```
jhughesbiot, line 22: else statement unnecessary as it is just setting the same variable with the existing value

bin/testrun
```diff
@@ -22,11 +22,16 @@ fi
...
-cd $TESTRUNPATH
+FILE_PATH=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
 
-# Create device folder if it doesn't exist
-mkdir -p local/devices
+# Check if Testrun was installed as a package
+if [ $FILE_PATH == "/usr/bin" ] ; then
+  export TESTRUN_PATH="/usr/local/testrun"
+else
+  export TESTRUN_PATH=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )/..
```
jhughesbiot, line 31: This doesn't resolve the root testrun directory as expected, should be updated to:

`export TESTRUN_PATH=$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")"/.. && pwd)`

cmd/install
jboddey: The docs are shortly going to be updated to state installing to a custom directory will require the following command: sudo TESTRUN_DIR=custom/dir cmd/install. This will allow using the package installer and custom directory

bin/testrun
jboddey: Does this not work for you? It works for me currently

jhughesbiot: How are you running testrun?  I run it from the testrun folder with the command sudo bin/testrun which resolves the TESTRUN_PATH incorrectly but the updated code works as expected.

cmd/install
jhughesbiot: Could we make this a command line arg instead?  Something like --install-dir.  This might be a cleaner option and then update this script to accept the args as necessary.

jboddey: I have just added the option for -l to instal locally

bin/testrun
jboddey: sudo bin/testrun is working for me after local install

jboddey: Both methods seem to work. I have used your suggestion in case that solves any issues on your end.

jhughesbiot (approved):

PR #293, opened 2024-02-23
Prevent a module from failing

jboddey:


jboddey:
This is going to be released in a hotfix

PR #305, opened 2024-03-01
Set test status to Error for In-Progress tests when cancelling Testrun

jboddey:


jhughesbiot:
Works as expected but is "error" the best descriptive state for this type of action?  Should it be stopped, killed or interrupted?

jboddey:
We have gone with Error for now as this is supported on the UI, other statuses wouldn't be rendered correctly. If we decide that stopped/killed etc is the better choice we will add support for this in the UI at a later date.

KrystsinaT (commented):
Verified. The test status "Error" is returned  for tests which were not completed while cancelling the test attempt.

jhughesbiot (approved):

PR #306, opened 2024-03-01
Adds version analytics event

sofyakurilova:


modules/ui/src/app/components/version/version.component.spec.ts
```diff
@@ -29,7 +29,7 @@ import { MatDialogRef } from '@angular/material/dialog';
 import { DeviceFormComponent } from '../../pages/devices/components/device-form/device-form.component';
 import { ConsentDialogComponent } from './consent-dialog/consent-dialog.component';
 
-describe('VersionComponent', () => {
+fdescribe('VersionComponent', () => {
```
OlgaMardvilko, line 32: Please change back to `describe`

OlgaMardvilko (changes requested):

PR #313, opened 2024-03-07
327161779: (fix) change get status with finished data to display a proper status on status bar

OlgaMardvilko:
### Overview

**Ticket:** 327161779
**fix:** change get status with finished data to display a proper status on status bar

### Screenshots after changes (with finished property of data)

image

image

image

### Screenshots before changes (with finished property of data)

image

image

image

image

### DIFF
--- a/modules/ui/src/app/pages/testrun/components/progress-status-card/progress-status-card.component.spec.ts
+++ b/modules/ui/src/app/pages/testrun/components/progress-status-card/progress-status-card.component.spec.ts
@@ -171,6 +171,17 @@ describe('ProgressStatusCardComponent', () => {
 
         expect(result).toEqual(expectedResult);
       });
+
+      it('should return test status "Monitoring" if finished with status "Monitoring"', () => {
+        const MOCK_MONITORING_FINISHED_DATA = {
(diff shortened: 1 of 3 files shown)


jboddey:
I'm not sure the changes have quite fixed the issue.

What has caused the below scenario to occur?
image

With the below image, the test status should say 'Cancelling' (if the cancel button has just been pressed)
image

And with this image, should it not say Incomplete instead of Complete?
image

OlgaMardvilko:
> I'm not sure the changes have quite fixed the issue.
> 
> What has caused the below scenario to occur? image
> 
> With the below image, the test status should say 'Cancelling' (if the cancel button has just been pressed) image
> 
> And with this image, should it not say Incomplete instead of Complete? image

Hi @jboddey 
There is the ticket where the problem is described: 327161779

In short, we previous used the logic that define the Test status value based on the finished property of data (Complete = has finished value). But, in the bug you can see some cases where we have this finished value, but statuses related more to progressing.
So, to be more accurate on UI we decided to fix it on UI as well. 
If you have more questions, let discuss.

jboddey (approved):

PR #314, opened 2024-03-07
Improve design for module pages

jboddey:


jboddey:
I have attached the new design: 
[report.pdf](https://github.com/google/testrun/files/14536504/report.pdf)

jhughesbiot:
Why is the nmap module referred to services?

jhughesbiot (approved):

jboddey:
> Why is the nmap module referred to services?

nmap will probably not mean a lot to users of Testrun, services is also the name used in the test plan

PR #320, opened 2024-03-11
Fix report removal from session

jboddey:


jhughesbiot:
Why delete this function?  Why not just change the default behavior to only cleanup if a max is set?

jboddey:
I'm just cautious that we have set a default limit in the config so a lot of users will have one set without knowing. This is a temporary fix until we can think of a better solution if needed. The current issue we have is that deleted reports still show up in the UI until Testrun is restarted so a bug is caused on the UI.

jhughesbiot (approved):

PR #326, opened 2024-03-13
Merge dev into main (Release v1.2)

jboddey:


modules/test/base/base.Dockerfile
```diff
@@ -50,8 +50,12 @@ ARG CONTAINER_PROTO_DIR=testrun/python/src/grpc_server/proto
 COPY $NET_MODULE_DIR/dhcp-1/$NET_MODULE_PROTO_DIR $CONTAINER_PROTO_DIR/dhcp1/
 COPY $NET_MODULE_DIR/dhcp-2/$NET_MODULE_PROTO_DIR $CONTAINER_PROTO_DIR/dhcp2/
 
+# Copy the cached version of oui.txt incase the download fails
+RUN mkdir -p /usr/local/etc
+COPY $MODULE_DIR/usr/local/etc/oui.txt /usr/local/etc/oui.txt
+
 # Update the oui.txt file from ieee
-RUN wget https://standards-oui.ieee.org/oui/oui.txt -P /usr/local/etc/
+RUN wget https://standards-oui.ieee.org/oui/oui.txt -O -P /usr/local/etc/ || echo "Unable to update the MAC OUI database"
```
jhughesbiot, line 58: Bad parameter -O added which breaks the oui.txt download causing oui test to fail.  Need to remove it

jhughesbiot: Cached file should also be updated as this wouldn't have been noticed if the cache file was up to date.

jhughesbiot: Check your build logs, you should see that the download failed to write the file and is using the cache.  You won't see the failure if you happen to be using a device within the cached file which is out of date.

jhughesbiot: https://github.com/google/testrun/pull/365

modules/test/tls/python/requirements.txt
```diff
@@ -1,2 +1,4 @@
 cryptography
-pyOpenSSL
\ No newline at end of file
+pyOpenSSL
+pyshark
+requests
```
jhughesbiot, line 4: Need to pin requirements for tls module.  Package updates can cause python crash if they get out of sync.
https://github.com/google/testrun/pull/366

travellingmonk (approved):

jhughesbiot (approved):

PR #329, opened 2024-03-13
Fix style for helperbird

sofyakurilova:
After changes
4Lpen3dy8X9vGBF

C9ky3sax4ob4hcg

8bwaRfJC9mCRUe2

9PEtecotRLWFWga


modules/ui/src/styles.scss
```diff
@@ -247,3 +247,21 @@ body:has(app-callout .info):has(app-callout .error)
     opacity: var(--mat-text-button-focus-state-layer-opacity);
   }
 }
+
+.device-item-with-actions[class*='helperbird'],
+.progress-card[class*='helperbird'] {
+  width: max-content !important;
```
OlgaMardvilko, line 253: I am not sure if it a right way to fix the issues with different type of fonts that people can choose and configure. The class `class*='helperbird'` only the class that one tools set to check it.

Issue #317, opened 2024-03-07
TLS client gives non-compliant if endpoint cannot be reached


jboddey:
**Describe the bug**
The TLS client test returns a non-compliant result if the device attempts to reach a device securely on port 443 but the device is not reachable. This is because no TLS handshake can occur, and only outbound tcp packets are seen.

**To Reproduce**
1. Configure device to communicate with a fake external web service on port 443
2. Run TLS module

**Expected behavior**
The test module should check if the outbound connection was successful

**Error logs**
[If applicable, provide a log from https://gist.github.com/](https://gist.github.com/jboddey/ed6f808c1339d6a58f32b9448348c937)
![image](https://github.com/google/testrun/assets/7399056/8863363e-e667-4246-a2d8-efd83f6c8365)

jhughesbiot:
I can't confirm the issue based on that capture file.  Even if a handshake wasn't completed, it would still have a client hello packet somewhere if it was a TLS connection.  Port 443 alone doesn't mean anything, it's just a default port.  Is this scenario described, how was the device setup to make a TLS connection to this fake server?

jboddey:
The protocol found was LON/IP. The TLS module only check for connections to the devices subnet locally which is why it has received a non-compliant.

jboddey:
This has been resolved in dev. Will be released in v1.2 next week.

PR #352, opened 2024-03-25
Add private ip subnet filter for non-tls traffic

jhughesbiot:
Allows all unencrypted traffic to private subnets initiated from the device.  Attempts to address https://github.com/google/testrun/issues/317


jboddey (approved):
Tested and working

PR #364, opened 2024-04-02
Returns the currently installed version if Testrun fails to fetch the latest version

jboddey:


framework/python/src/api/api.py
```diff
@@ -243,8 +247,35 @@ async def stop_test_run(self):
...
+      )
+
+    self._test_run.shutdown()
+    os.kill(os.getpid(), signal.SIGTERM)
+
   async def get_version(self, response: Response):
+
+    # Add defaults
     json_response = {}
+    json_response["installed_version"] = "v1.0"
```
jhughesbiot, line 274: Hard coding doesn't give correct results.  Should resolve current version from the session:

`json_response["installed_version"] = self._test_run.get_session().get_version()`

jboddey: This is done shortly after in the code. The idea here was to add a default fallback value in case resolving the current version fails for whatever reason.

jhughesbiot: If that's the case, we need  report some sort of unresolved version if this fails or needs a different reliable way to have a default installed version reported.  Hard coding a wrong version doesn't solve this, just opens the door for bad version reporting.

jhughesbiot:
Found an issue where the _load_version method in session was not correctly detecting failures and would essentially blank out the version if it failed.  Updated the method and added a default version parameter and set the self._version only in this method instead of in the constructor and this method.

jhughesbiot (approved):

PR #372, opened 2024-04-04
Fix bug with deleting reports

jboddey:


KrystsinaT (commented):
Verified. The report is deleted and not shown after refreshing the list.

travellingmonk (approved):
report does not appear once deleted

PR #374, opened 2024-04-04
Rename nmap module to services

jhughesbiot:


jboddey:
The only issue I can see with this change, is that every existing device that has the nmap module enabled it has now been disabled. Could we add logic to update every device config when Testrun launches or when config is loaded?

jhughesbiot:
> The only issue I can see with this change, is that every existing device that has the nmap module enabled it has now been disabled. Could we add logic to update every device config when Testrun launches or when config is loaded?

Unclear on this issue, all modules are enabled by default unless the config file for the module or the device is set to disabled.  I would expect this to actually be the reverse where nmap modules set as disabled would be enabled.

jboddey (approved):

PR #375, opened 2024-04-05
Extend timeout when testrun is stopping

sofyakurilova:
+ extend timeout for next requests until stop


OlgaMardvilko (commented):
Should we cover this changes with unit tests?

PR #376, opened 2024-04-05
monitor device disconnect

jhughesbiot:
- Monitor and cancel testrun if device interface is disconnected during monitor phase
 - Shutdown network if cancelled during monitor phase


jboddey:
Just an issue with pylint to look at

jboddey (approved):

PR #379, opened 2024-04-08
Create SECURITY.md

PranithChowdary:


jboddey:
Going to close this for now. The current policy for security updates with Testrun is that only the latest version is supported. We are not going to maintain older versions.

PR #383, opened 2024-04-09
Fixes defects in the PDF report if a long device name is specified

jboddey:
See attached PDF

[2024-04-09T13_09_48.pdf](https://github.com/google/testrun/files/14918718/2024-04-09T13_09_48.pdf)

### DIFF
--- a/framework/python/src/common/testreport.py
+++ b/framework/python/src/common/testreport.py
@@ -279,7 +279,7 @@ class TestReport():
         # Add appropriate content size for each data row
         # update if CSS changes for this element
         elif '<tr>' in line and data_rows_active:
-          content_size += 40.667
+          content_size += 42
 
         # If the current line is within the content size limit
         # we'll add it to this page, otherweise, we'll put it on the next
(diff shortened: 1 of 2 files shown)


framework/python/src/common/testreport.py
```diff
@@ -23,7 +23,7 @@
 
 DATE_TIME_FORMAT = '%Y-%m-%d %H:%M:%S'
 RESOURCES_DIR = 'resources/report'
-TESTS_FIRST_PAGE = 12
+TESTS_FIRST_PAGE = 9
```
jhughesbiot, line 26: This should be dynamic depending on the size of the header.  Making it 9 leaves a lot of dead space with any manufacturer not having too long of a name problem.

jhughesbiot (commented):
Spacing for longer names cuts off multi-page module reports.  Needs to dynamically account for this extra space needed for the header.

image

KrystsinaT (commented):
Verified. The Post request to create a device with more than 28 characters results in 400 {"error":"Invalid request received"}.

framework/python/src/common/testreport.py
jboddey: It has now been set to 11 which should be enough space for a device with a long name, without causing much white space for short names.

jhughesbiot:
> Spacing for longer names cuts off multi-page module reports. Needs to dynamically account for this extra space needed for the header.
> 
> image

Updated the row size to account for the updated CSS changes so this should be resolved now.

jhughesbiot (approved):

PR #387, opened 2024-04-10
Fix modbus test

jhughesbiot:
Modbus result check returns wrong type crashing protocol module when modbus test is compliant.  This fixes the result check to return the expected boolean type

### DIFF
--- a/cmd/package
+++ b/cmd/package
@@ -17,7 +17,7 @@
 # Creates a package for Testrun
 
 MAKE_SRC_DIR=make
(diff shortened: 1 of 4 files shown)


jboddey:
Can we recreate this PR on main and release this as a hotfix? As version v1.2.1?

jhughesbiot:
Replaced by https://github.com/google/testrun/pull/389

Should delete and merge from main once above PR is merged

PR #389, opened 2024-04-11
Hotfix 1.2.1

jhughesbiot:
Fix modbus test crash when compliant

### DIFF
--- a/cmd/package
+++ b/cmd/package
@@ -56,4 +56,4 @@ cp -r {framework,modules} $MAKE_SRC_DIR/usr/local/testrun
 dpkg-deb --build --root-owner-group make
 
 # Rename the .deb file
-mv make.deb testrun_1-2_amd64.deb
+mv make.deb testrun_1-2-1_amd64.deb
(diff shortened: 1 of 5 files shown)


jboddey:
Also, whilst I was looking at this, I found that there should be extra logic in the test module base code where checking what is returned by each test. Where we check if the result is equal to error, we need to add an else statement. Could you add this whilst I'm away?

jhughesbiot:
> Also, whilst I was looking at this, I found that there should be extra logic in the test module base code where checking what is returned by each test. Where we check if the result is equal to error, we need to add an else statement. Could you add this whilst I'm away?

Do we just want to make this an else statement? I don't actually see any benefit for specifically checking for error

jboddey:
> > Also, whilst I was looking at this, I found that there should be extra logic in the test module base code where checking what is returned by each test. Where we check if the result is equal to error, we need to add an else statement. Could you add this whilst I'm away?
> 
> Do we just want to make this an else statement? I don't actually see any benefit for specifically checking for error

Yes exactly, the else will then return 'Error'

jhughesbiot:
Updated to handle unknown result values

jboddey (approved):

PR #390, opened 2024-04-11
Support certificate names with spaces in TLS test

gabrielandrade2:
I noticed that the TLS test will fail if the Root CA file placed in `local/root_certs` has a space in its name.

The filename parameters are not being properly handled when passed from `tls_utils.py` to `check_cert_signature.sh`, thus the openssl call in the bash script returns `No such file or directory`.

As the `get_started.md` file doesn't make any restriction to the file name, it can be misleading not supporting spaces: 
> NOTE: Place your local CA certificate in local/root_certs (any name with a .crt extension) to perform TLS server tests

In this PR, I basically added double quotes to parameters that can be a file.
I've also included logs to the response of the bash commands for easier debugging.

I only touched the TLS module, the same issue can still happen in other modules

### DIFF
--- a/modules/test/tls/python/src/tls_util.py
+++ b/modules/test/tls/python/src/tls_util.py
@@ -188,9 +188,10 @@ class TLSUtil():
         # Create the file path
         root_cert_path = os.path.join(self._root_certs_dir, root_cert)
         LOGGER.info('Checking root cert: ' + str(root_cert_path))
-        args = f'{root_cert_path} {device_cert_path}'
+        args = f'"{root_cert_path}" "{device_cert_path}"'
         command = f'{bin_file} {args}'
         response = util.run_command(command)
+        LOGGER.debug(response)
         if f'{device_cert_path}: OK' in str(response):
           LOGGER.info('Device signed by cert:' + root_cert)
           return True, root_cert_path
@@ -257,9 +258,10 @@ class TLSUtil():
 
     # Use openssl script to validate the combined certificate
     # against the available trusted CA's
-    args = f'{intermediate_cert_path} {combined_cert_path}'
+    args = f'"{intermediate_cert_path}" "{combined_cert_path}"'
     command = f'{bin_file} {args}'
     response = util.run_command(command)
+    LOGGER.debug(response)
     return combined_cert_name + ': OK' in str(response)
 
   def get_ca_issuer(self, certificate):
@@ -407,9 +409,10 @@ class TLSUtil():
 
   def get_ciphers(self, capture_file, dst_ip, dst_port):
     bin_file = self._bin_dir + '/get_ciphers.sh'
-    args = f'{capture_file} {dst_ip} {dst_port}'
+    args = f'"{capture_file}" {dst_ip} {dst_port}'
     command = f'{bin_file} {args}'
     response = util.run_command(command)
+    LOGGER.debug(response)
     ciphers = response[0].split('\n')
     return ciphers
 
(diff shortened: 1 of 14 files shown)


jboddey:
@jhughesbiot could you review this PR?

jhughesbiot:
I wrote a unit test around this that I would like you to incorporate into this PR before merging so we can prevent any regression issues.

Please add this test to the testing/unit/tls/tls_module_test.py file:

```
  def tls_module_ca_cert_spaces_test(self):
    print('\tls_module_ca_cert_spaces_test')
    # Make a tmp folder to make a differnt CA directory
    tmp_dir = os.path.join(TEST_FILES_DIR,'tmp')
    if os.path.exists(tmp_dir):
      shutil.rmtree(tmp_dir)
    os.makedirs(tmp_dir, exist_ok=True)
    # Move and rename the TestRun CA root with spaces
    ca_file = os.path.join(ROOT_CERTS_DIR,'Testrun_CA_Root.crt')
    ca_file_with_spaces = os.path.join(tmp_dir,'Testrun CA Root.crt')
    shutil.copy(ca_file,ca_file_with_spaces)

    cert_path = os.path.join(CERT_DIR,'device_cert_local.crt')
    log = logger.get_logger('unit_test_' + MODULE)
    log.setLevel(logging.DEBUG)
    tls_util = TLSUtil(log,
                       bin_dir='modules/test/tls/bin',
                       cert_out_dir=OUTPUT_DIR,
                       root_certs_dir=tmp_dir)

    cert_valid = tls_util.validate_local_ca_signature(device_cert_path=cert_path)
    self.assertEqual(cert_valid[0], True)
```

You will also need add the following imports:

```
import shutil
import logging
```

And then add it to the test order:

```
  # Test signature validation  methods
  suite.addTest(TLSModuleTest('tls_module_trusted_ca_cert_chain_test'))
  suite.addTest(TLSModuleTest('tls_module_local_ca_cert_test'))
  suite.addTest(TLSModuleTest('tls_module_ca_cert_spaces_test'))
```

You will also need to exclude the tmp directory created for this test in the .gitignore file

> testing/unit_test/temp/
> testing/unit/dns/output/
> testing/unit/nmap/output/
> testing/unit/ntp/output/
> testing/unit/tls/output/
> **testing/unit/tls/tmp/**
> testing/unit/report/output

modules/test/tls/python/src/tls_util.py
```diff
@@ -188,9 +188,10 @@ def validate_local_ca_signature(self, device_cert_path):
         # Create the file path
         root_cert_path = os.path.join(self._root_certs_dir, root_cert)
         LOGGER.info('Checking root cert: ' + str(root_cert_path))
-        args = f'{root_cert_path} {device_cert_path}'
+        args = f'"{root_cert_path}" "{device_cert_path}"'
         command = f'{bin_file} {args}'
         response = util.run_command(command)
+        LOGGER.debug(response)
```
jhughesbiot, line 194: The response object for this method returns a tuple so doesn't lend itself to a very user friendly logging 

Error Example:
> ('ROOT: testing/unit/tls/tmp/Testrun\nDEVICE_CERT: CA', b'Error loading file testing/unit/tls/tmp/Testrun\n')

Success Example:

> ('ROOT: testing/unit/tls/tmp/Testrun CA Root.crt\nDEVICE_CERT: testing/unit/tls/certs/device_cert_local.crt\ntesting/unit/tls/certs/device_cert_local.crt: OK', b'')

Adding logging to the result should check for error before logging and only log success results since error logging is already a built in function to the underlying run_command function and would cause duplicate logs.
  ```
stdout, stderr = process.communicate()
  if process.returncode != 0 and output:
    err_msg = f'{stderr.strip()}. Code: {process.returncode}'
    LOGGER.error('Command Failed: ' + cmd)
    LOGGER.error('Error: ' + err_msg)
```

See example here of error checking of run_cmd:
https://github.com/google/testrun/blob/dev/framework/python/src/common/session.py#L143-L154

gabrielandrade2: Oh, I didn't notice that error log before.
Perhaps having the debugs logs directly into the 'run_command'  functions instead is a better idea.

gabrielandrade2:
@jhughesbiot I've added the test and also reworked the logging approach.

I moved the log call inside `run_command` and avoid log the `stderr` output two time.

gabrielandrade2:
@jhughesbiot Oh, my bad. I didn't realize there was more than one `run_command`. I've changed all of them now.

I also committed the changes on the gitignore file I forgot to include last time.

jhughesbiot (approved):

PR #396, opened 2024-04-15
Add feature not detected test result

jboddey:


jhughesbiot:
Looks like this only adds the option to the report but no way to generate this result, is this intentional? Or is there more work expected on this PR to update the tests to allow this option?

Also, I see a lot of pylint fixes here, should modify adjust the pylint test to match the updates.

modules/test/ntp/python/src/ntp_module.py
```diff
@@ -272,11 +272,11 @@ def _ntp_network_ntp_dhcp(self):
         result = False, ('Device sent NTP request to DHCP provided ' +
                          'server and non-DHCP provided server')
       elif ntp_to_remote:
-        result = False, 'Device sent NTP request to non-DHCP provided server'
+        result = 'Feature Not Present', 'Device sent NTP request to non-DHCP provided server'
       elif ntp_to_local:
         result = True, 'Device sent NTP request to DHCP provided server'
     else:
-      result = False, 'Device has not sent any NTP requests'
+      result = 'Feature Not Present', 'Device has not sent any NTP requests'
```
jhughesbiot, line 279: Replacing the failure with a feature not present is not correct. These are failure states.

modules/test/protocol/python/src/protocol_modbus.py
```diff
@@ -267,7 +267,7 @@ def validate_device(self):
       compliant = (holding_reg is not None or input_reg is not None
                    or coils is not None or discrete_inputs is not None)
     else:
-      compliant = False
+      compliant = 'Feature Not Present'
```
jhughesbiot, line 270: Replacing the failure with a feature not present is not correct. There is no way to determine if a device supports Modbus feature if not discovered, this needs further configuration.

modules/test/tls/python/src/tls_module.py
```diff
@@ -280,7 +280,7 @@ def _validate_tls_client(self, client_ip, tls_version):
 
     # Generate results based on the state
     result_message = 'No outbound connections were found.'
-    result_state = None
+    result_state = 'Feature Not Present'
```
jhughesbiot, line 283: Replacing the None with a feature not present is not correct. There is no way to determine if a device supports TLS client feature if not communicating during test, this needs further configuration.

jhughesbiot (commented):
Adding feature not present to test results over failures is not a valid way to determine this.  A device must define whether or not it supports a feature so it can be tested accordingly.  A test needs to generate pass/fail result for a test.  It can only get a feature not present result if the device configuration has been set to override failures because it doesn't support that feature.

jboddey:
> Adding feature not present to test results over failures is not a valid way to determine this. A device must define whether or not it supports a feature so it can be tested accordingly. A test needs to generate pass/fail result for a test. It can only get a feature not present result if the device configuration has been set to override failures because it doesn't support that feature. However, we will rephrase Feature Not Present to something else, like Feature Not Detected.

As mentioned, we have decided to continue with Testrun determining whether a device currently has the feature enabled. We do not care if a device has a feature or does not have a feature, it is only the current configuration that we are testing. Whether a feature is disabled or not will be picked up in the ideal config.

modules/test/dns/python/src/dns_module.py
```diff
@@ -213,46 +212,46 @@ def _dns_network_from_dhcp(self):
 
     if dns_packets_local or dns_packets_not_local:
       if dns_packets_not_local:
-        result = False, 'DNS traffic detected to non-DHCP provided server'
+        # Feature Not Detected
+        result = ('Feature Not Detected',
+                  'DNS traffic detected to non-DHCP provided server')
       else:
```
jhughesbiot, line 218: I believe this is still a failure scenario, DNS traffic is detected but not using the DHCP provided DNS servers.  No DNS traffic is captured as a feature not present.

jboddey: I would suggest the opposite. If the device is sending no DNS at all, then I would argue that this could be either Feature Not Detected (because it has not detected the feature of using local DNS servers) or Non-Compliant (because no DNS is considered Non Compliant. However, if the device is sending DNS, but to other DNS servers, this must be a Feature Not Detected as DNS is compliant and the feature of using local DNS servers is not detected (configured or present) on the device.

modules/test/protocol/python/src/protocol_bacnet.py
```diff
@@ -54,23 +57,21 @@ def discover(self, local_ip=None):
...
     self.discover(local_ip + '/24')
-    LOGGER.info('BACnet Devices Found: ' + str(len(self.devices)))
+    LOGGER.info('BACnet devices found: ' + str(len(self.devices)))
     if len(self.devices) > 0:
-      # Load a fail result initially and pass only
-      # if we can validate it's the right device responding
-      result = False, ('Could not confirm discovered BACnet device is the ' +
-                       'same as device being tested')
+      result = ('Feature Not Detected',
+                'Device did not respond to BACnet discovery')
```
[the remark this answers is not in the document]
jboddey: The reason I left this as Feature Not Detected is because I can see a use case (either now or in the future), where we allow users to connect the device testing interface to a switch with multiple devices. If there are other BACnet devices on that switch, that is fine - the user may have configured the device under test to communicate with peer devices using other protocols.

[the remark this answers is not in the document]
jhughesbiot: I think until we have that option, we can't change the results of this test.  Once we have that ability we should re-evaluate how this is determined to account for multiple device connections as that scenario will be a sprawling change to almost all tests.

modules/test/dns/python/src/dns_module.py
jhughesbiot: I think you  might be conflating the two DNS tests into one when making those determinations.  Hostname resolution is what should only care that DNS was used to determine compliant/non-compliant or feature not detected.  DNS local is very clearly defined by the expected behavior:

        "test_description": "Verify the device allows for a DNS server to be entered automatically", 
        "expected_behavior": "The device sends DNS requests to the DNS server provided by the DHCP server",

If a device is using DNS to a remote DNS server, that's a clear failure, not a feature not detected.  This can only be feature not detected if no DNS is detected.  Otherwise, if DNS is present and not to the local, it fails because it is using DNS just not as defined.

modules/test/protocol/python/src/protocol_bacnet.py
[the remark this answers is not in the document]
jboddey: Ok, I have reverted this one back to non-compliant.

modules/test/dns/python/src/dns_module.py
jboddey: This does not line up with the definition of a recommended, required if applicable or optional test. These tests should never return Non-Compliant because that indicates that the device under test is a non-compliant device, which it is not.

jhughesbiot: A test result should be independent of the requirement of the test.  The overall compliance of the device will take into account the requirement of the individual tests and the result to calculate it's final score.  In this case, this is a roadmap/recommended test which would still pass on compliance if the test fails while indicating it is not complaint to this specific test.  I don't see that as an issue.

jhughesbiot (approved):

PR #409, opened 2024-04-24
Remove report count limit

jboddey:


jhughesbiot:
Where is the request coming from for the removal of this limit? I don't like the idea of removing useful features without justification.

jboddey:
Requested by the UI team. General user expectations would be that all reports are available. There does not seem to be a reason why we would delete reports now that users have the ability to delete reports.

jhughesbiot:
This PR is now superseded by https://github.com/google/testrun/pull/423

Change behavior to keep all reports instead of removing limit option

PR #411, opened 2024-04-24
Add steps to resolve to PDF report

jboddey:


modules/test/ntp/python/src/ntp_module.py
```diff
@@ -191,6 +192,8 @@ def extract_ntp_data(self):
               'Timestamp': float(packet.time),
           })
 
+          print("Appending " + str(packet.time))
+
```
jhughesbiot, line 196: Debugging statement? Remove or replace with logger if necessary.

jhughesbiot:
Does not handle multi-page results.

image

jhughesbiot:
Shows blank results page if no failures are present with steps.  Update to generate_steps_to_resolve to wrap the page contents only if they are non-empty at the same time as implementing a multipage solution.

image

jboddey:
> Does not handle multi-page results.
> 
> image

This should be resolved now. Could you give it another try?

jhughesbiot:
I updated the unit tests with both compliant and non-compliant outputs.  Should be able to run these for subsequent report changes to quickly generate reports for inspection on these type of issues.  Everything seems good now using these tests to confirm.

jhughesbiot (approved):

PR #441, opened 2024-05-09
Fix alignment with steps to resolve in PDF report

jboddey:
See attached.

[report.pdf](https://github.com/google/testrun/files/15259763/report.pdf)


KrystsinaT (commented):
Verified. The steps are aligned. Thank you!

PR #449, opened 2024-05-11
Fix modify devices test

jboddey:


framework/python/src/net_orc/ip_control.py
```diff
@@ -42,10 +42,12 @@ def add_namespace(self, namespace):
       return success
 
   def check_interface_status(self, interface_name):
-    output = util.run_command(cmd=f'ip link show {interface_name}',output=True)
+    output = util.run_command(cmd=f'ip link show {interface_name}', output=True)
     if 'state DOWN ' in output[0]:
+      LOGGER.debug('Device interface is down')
       return False
     else:
+      LOGGER.debug('Device interface is up')
```
jhughesbiot, line 50: Adding this logging is unnecessary and will clog up the main log output since it is called in a loop during the monitor process.  Should leave any logging up the caller.

jhughesbiot (approved):

PR #453, opened 2024-05-14
Reduce locations of Testrun version

jboddey:
Removes multiple declarations of the Testrun version number. The version should now be specified in make/DEBIAN/control only, and occasionally updated in testrun.py as a fallback.


framework/python/src/core/testrun.py
```diff
@@ -60,7 +60,8 @@
 DEVICE_TEST_MODULES = 'test_modules'
 MAX_DEVICE_REPORTS_KEY = 'max_device_reports'
 
-VERSION = '1.2.2'
+# Does not need to be updated all the time, just a fallback
+VERSION = '1.3-alpha'
```
jhughesbiot, line 65: Is there a reason we can't just resolve this from the control file directly instead of potentially having a bad fallback value?

framework/python/src/common/testreport.py
```diff
@@ -134,6 +136,12 @@ def to_json(self):
 
   def from_json(self, json_file):
 
+    # Version added in v1.3-alpha
+    if 'testrun' in json_file and 'version' in json_file['testrun']:
+      self._version = json_file['testrun']['version']
+    else:
+      self._version = '1.3-alpha'
+
```
jhughesbiot, line 144: Could we pass the session or version into the constructor for the TestReport object to set a default?

jboddey: The version here is the version which was used when the report was executed, not the current version of Testrun. I have replaced the fallback version with 'Unknown'

framework/python/src/core/testrun.py
jboddey: Removed the fallback

jhughesbiot (approved):

PR #457, opened 2024-05-15
Generate SSL certificate for Testrun UI

jboddey:
The certificate is not automatically trusted for Chrome. The user must manually import the certificate by following this guide: https://docs.vmware.com/en/VMware-Adapter-for-SAP-Landscape-Management/2.1.0/Installation-and-Administration-Guide-for-VLA-Administrators/GUID-D60F08AD-6E54-4959-A272-458D08B8B038.html

Documentation is yet to be added to this PR.


jboddey:
Any opinions on the implementation of this @jhughesbiot?

jhughesbiot:
@jboddey Fixed a few issues so this should be good to go.
One thing I noticed is that the auto trusting of the cert will only work for connections directly on the same machine and won't work for remote connections.  Similar to the MDNS registered domain so i'm not sure even trusting it, without a proper MDNS network configuration if this will solve browser auth errors but it does at least add HTTPS.

jhughesbiot:
@jboddey I can't seem to get this to actually work anymore, UI module keeps crashing.

jboddey:
This no longer works because the UI connects to the ws server using an insecure port (so it doesn't load). In order to continue, we would need to add TLS to the web sockets server.

PR #475, opened 2024-05-23
Add get profiles format endpoint

jboddey:


framework/python/src/common/session.py
```diff
@@ -189,7 +192,7 @@ def _load_version(self, default_version):
       version = version_cmd[0]
       self._version = version
     else:
-      self._version = default_version
+      self._version = '?'
```
jhughesbiot, line 195: Should be changed to unknown to match report values when the version is not resolved.

jhughesbiot (approved):

PR #497, opened 2024-06-05
Adds download zip window

sofyakurilova:
RddLCTzeVzRqVKb

video after changes 
https://screencast.googleplex.com/cast/NDUyNDM2NzkyODA5ODgxNnw3OTdlNThlYy0zYw

### DIFF
--- a/modules/ui/src/app/components/download-report-zip/download-report-zip.component.spec.ts
+++ b/modules/ui/src/app/components/download-report-zip/download-report-zip.component.spec.ts
@@ -13,75 +13,153 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import { ComponentFixture, TestBed } from '@angular/core/testing';
+import {
+  ComponentFixture,
+  fakeAsync,
+  TestBed,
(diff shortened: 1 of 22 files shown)


modules/ui/src/app/components/download-report-zip/download-report-zip.component.spec.ts
```diff
@@ -13,75 +13,153 @@
...
-      });
+      expect(openSpy).toHaveBeenCalled();
+
+      openSpy.calls.reset();
     });
   });
 });
+
+@Component({
+  selector: 'app-fake-risk-assessment-component',
```
OlgaMardvilko, line 162: Could you explain why do we need it here?

OlgaMardvilko (approved):

modules/ui/src/app/components/download-report-zip/download-report-zip.component.spec.ts
sofyakurilova: Navigator can't navigate without providing component in router

PR #498, opened 2024-06-05
Fix issue with checking for error result

jboddey:


framework/python/src/common/testreport.py
```diff
@@ -430,7 +430,7 @@ def generate_results(self, json_data, page_num):
 
     successful_tests = 0
     for test in self._results:
-      if test['result'] != 'Error':
+      if test.result != 'Error':
```
jhughesbiot, line 433: This should be using the json data provided instead of the results var so that it is validating the same data that is used in the rest of the method.  Additional issues in the html as there is no len parameter for an in.  See below code that should correct this.

```
    successful_tests = 0
    for test in json_data['tests']['results']:
      if test['result'] != 'Error':
        successful_tests += 1

    result_list = f'''
      <div class="result-list">
        <h3>Results List <small>({successful_tests}/{self._total_tests})</small></h3>
        <div class="result-line" style="margin-top: 10px;border-top-left-radius:4px;border-top-right-radius:4px;">
          <div class="result-list-header-label" style="left: .1in">Name</div>
          <div class="result-list-header-label" style="left: 2.8in">Description</div>
          <div class="result-list-header-label" style="left: 7.3in">Result</div>
        </div>'''
```

jboddey: Good catch, I have updated.

jhughesbiot: Need to update the html since it's still trying to access the len function of an int, which will throw an error.

This:
`<h3>Results List <small>({len(successful_tests)}/{self._total_tests})</small></h3>`

Needs to be
`<h3>Results List <small>({successful_tests}/{self._total_tests})</small></h3>`

jhughesbiot (approved):

PR #517, opened 2024-06-13
The Downlaod ZIP action can not be performed using the keyboard

sofyakurilova:
Stop propagating event to eliminate the error
"+" fix tooltip
BD6pc5LTD3LVc2R

### DIFF
--- a/framework/python/src/api/api.py
+++ b/framework/python/src/api/api.py
@@ -87,7 +87,8 @@ class Api:
     self._router.add_api_route("/report/{device_name}/{timestamp}",
                                self.get_report)
     self._router.add_api_route("/export/{device_name}/{timestamp}",
-                               self.get_results)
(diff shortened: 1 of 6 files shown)


jboddey:
I have fixed an issue with the API where the export endpoint had not been updated to accept POST requests. However, the UI is trying to interpret the response as JSON instead of downloading a ZIP file @sofyakurilova

sofyakurilova:
> I have fixed an issue with the API where the export endpoint had not been updated to accept POST requests. However, the UI is trying to interpret the response as JSON instead of downloading a ZIP file @sofyakurilova

Fixed http://screencast/cast/NTUwNjI2MTYzMzIwNDIyNHxjN2MxY2FkNy0yOQ

jboddey (approved):

PR #519, opened 2024-06-17
Add required if applicable

jboddey:


modules/test/tls/python/src/tls_module.py
```diff
@@ -281,11 +281,13 @@ def _security_tls_v1_2_client(self):
...
       if results[0] is None:
         description = 'No outbound connections were found'
+        return 'Feature Not Detected', description
       elif results[0]:
         description = 'TLS 1.2 client connections valid'
+        return True, description
       else:
         description = 'TLS 1.2 client connections invalid'
-      return results[0], description,  results[1]
+        return False, description
```
jhughesbiot, line 290: Should update a result field and still only have on return statement to maintain the results[1], which is the test details required for the report.

```
      # Determine results and return proper messaging and details
      description = ''
      result = None
      if results[0] is None:
        description = 'No outbound connections were found'
        result = 'Feature Not Detected'
      elif results[0]:
        description = 'TLS 1.2 client connections valid'
        result = 'True'
      else:
        description = 'TLS 1.2 client connections invalid'
        result = 'False'
      return result, description,  results[1]
```

jhughesbiot (approved):

PR #522, opened 2024-06-17
Feature/risk profile

jhughesbiot:
- Move all risk profiles methods to risk profile class
- Validate risk profile on creation and update
- Add risk profile unit tests


framework/python/src/common/risk_profile.py
```diff
@@ -33,27 +33,29 @@ def __init__(self, profile_json=None, profile_format=None):
     if profile_json is None or profile_format is None:
       return
     self.name = profile_json['name']
-    self.created = datetime.now()
+    self.created = datetime.now().strftime('%Y-%m-%d')
     self.version = profile_json['version']
```
jhughesbiot, line 37: Keep this as a timestamp, no need to convert to string outside o to_json method.

framework/python/src/common/risk_profile.py
```diff
@@ -33,27 +33,29 @@ def __init__(self, profile_json=None, profile_format=None):
...
     self._validate(profile_json, profile_format)
     self._update_categories()
+    self._update_risk()
 
   # Load a profile without modifying the created date
   # but still validate the profile
   def load(self, profile_json, profile_format):
     self.name = profile_json['name']
-    self.created = datetime.strptime(profile_json['created'],
-                                     '%Y-%m-%d %H:%M:%S')
```
jhughesbiot: Decode as a timestamp object so it doesn't break other methods

framework/python/src/common/risk_profile.py
```diff
@@ -152,15 +165,16 @@ def _valid(self, profile_json, profile_format):
 
   def _expired(self):
     # Check expiry
-    created_date = self.created.timestamp()
+    created_date = datetime.strptime(
+      self.created, '%Y-%m-%d').timestamp()
```
jhughesbiot, line 169: See note above about keeping this as a timestamp object, no need to decode string when kept as proper timestamp object

framework/python/src/common/risk_profile.py
```diff
@@ -152,15 +165,16 @@ def _valid(self, profile_json, profile_format):
...
     today = datetime.now().timestamp()
     return created_date < (today - SECONDS_IN_YEAR)
 
   def to_json(self, pretty=False):
     json_dict = {
         'name': self.name,
         'version': self.version,
-        'created': str(self.created),
+        'created': self.created,
```
jhughesbiot, line 177: See above comments concerning timestamp vs string object

framework/python/src/common/risk_profile.py
jboddey: The issue with using a timestamp, is that when it is passed to the UI, it passes it as a full timestamp in ISO format but I was wanting to pass it to the UI as a string in a particular format.

jboddey (approved):

PR #534, opened 2024-06-21
Feature/ws

hitnik:
MQTT protocol support.  MQTT client with functionality for sending messages to topic subscribers.

### DIFF
--- a/framework/python/src/common/mqtt.py
+++ b/framework/python/src/common/mqtt.py
@@ -18,7 +18,7 @@ import paho.mqtt.client as mqtt_client
 from common import logger
 from common import util
 
-LOGGER = logger.get_logger("runner")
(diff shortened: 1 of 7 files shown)


framework/python/src/common/mqtt.py
```diff
@@ -0,0 +1,59 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""MQTT client"""
+import typing as t
+import paho.mqtt.client as mqtt_client
+from common import logger
+from common import util
+
+LOGGER = logger.get_logger("runner")
```
jboddey, line 21: Could we rename the logger to something like mqtt_client?

hitnik: Renamed

jboddey:
I believe this contains the files from PR #561, closing this one.

PR #563, opened 2024-06-29
Fix bug when saving draft profile

jboddey:
This pull requests looks to resolve the following bug:

When the user selects 'Save Draft' in the UI, the profile is saved as a valid profile instead of a draft profile. In the code changes, a check for the 'status' field is completed. If the status is draft, then the status of the risk profile is set to draft. Previously, the risk profile was evaluated for risk **before** the answers were associated with the risk profile so the risk was outdated. Now, the risk is checked after the profile has been fully updated.

Also, additional validation has been added to the backend to ensure answers are valid (correct length, and at least one option is chosen). A unit test is added to verify this


jhughesbiot:
I don't know what bug is being fixed here to confirm if it's been resolved.  Can you elaborate on the issue being reported and fixed?  Add to the unit tests to account for this so we know it won't c come back in regression testing.

jhughesbiot (approved):

framework/python/src/common/session.py
```diff
@@ -501,29 +501,15 @@ def update_profile(self, profile_json):
 
     else:
 
+      # Update the profile
+      risk_profile.update(profile_json)
```
jboddey, line 505: I get this error: TypeError: RiskProfile.update() missing 1 required positional argument: 'profile_format'

framework/python/src/common/risk_profile.py
```diff
@@ -70,6 +70,8 @@ def update(self, profile_json, profile_format):
     self.questions = new_profile.questions
     self.status = new_profile.status
 
+    self.update_risk(profile_format=profile_format)
+
```
jhughesbiot, line 74: Should be 

`self.risk = new_profile.risk `

since this will already be calculated, no need to do it again

jboddey: Changed

framework/python/src/common/session.py
jboddey: Resolved.

PR #573, opened 2024-07-04
Adds copy of risk profile

sofyakurilova:
Tests
BTpmUaPdCRPsPmd

Video after changes
http://screencast/cast/NjM0ODAxMzU1MDE3NDIwOHxlZDZlZWUwNi0yMA

### DIFF
--- a/modules/ui/src/app/pages/risk-assessment/risk-assessment.component.ts
+++ b/modules/ui/src/app/pages/risk-assessment/risk-assessment.component.ts
@@ -26,6 +26,7 @@ import { MatDialog } from '@angular/material/dialog';
 import { LiveAnnouncer } from '@angular/cdk/a11y';
 import { Profile } from '../../model/profile';
 import { Observable } from 'rxjs/internal/Observable';
+import { DeviceValidators } from '../devices/components/device-form/device.validators';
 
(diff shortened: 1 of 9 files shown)


modules/ui/src/app/pages/risk-assessment/risk-assessment.component.ts
```diff
@@ -60,6 +60,24 @@ export class RiskAssessmentComponent implements OnInit, OnDestroy {
...
+
+  getCopyOfProfile(profile: Profile): Profile {
+    const copyOfProfile = { ...profile };
+    copyOfProfile.name = this.getCopiedProfileName(profile.name);
+    return copyOfProfile;
+  }
+
+  private getCopiedProfileName(name: string): string {
+    name = `Copy of ${name}`;
+    if (name.length > 28) {
```
OlgaMardvilko, line 75: I think it will be better to use the variable here. If I got it correct this magic number goes from the validation rule for max length. So, in case this number changes we could update it only in one single place. Suggestion: use single variable for the same/depended logic.

OlgaMardvilko (changes requested):

OlgaMardvilko (approved):

PR #577, opened 2024-07-04
Remove profile from runtime once included in ZIP

jboddey:
Precondition: Run the test attempt till completion to get the report. Create 1 complete risk profile.

Open the Reports page
Click the button Download ZIP for the report generated in the precondition.
Select the risk profile and click Download
Open the Risk assessment page and remove the risk profile. The list should be empty
Open the Reports page and click the button Download ZIP again
Click Download anyway ( without risk profile)
Open the ZIP folder Actual result: the profile PDF is available inside the folder. Expected: the profile is not available.

### DIFF
--- a/framework/python/src/test_orc/test_orchestrator.py
+++ b/framework/python/src/test_orc/test_orchestrator.py
@@ -272,7 +272,8 @@ class TestOrchestrator:
         timestamp)
 
       # Define where to save the zip file
-      zip_location = os.path.join('/tmp/testrun',timestamp)
+      zip_location = os.path.join("/tmp/testrun",
+                                  timestamp)
 
       # Delete ZIP if it already exists
       if os.path.exists(zip_location + ".zip"):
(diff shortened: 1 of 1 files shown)


jhughesbiot:
What is the reasoning behind deleting the profile after zipping?  What issue is this resolving?

jboddey:
> What is the reasoning behind deleting the profile after zipping? What issue is this resolving?

If you delete a profile that has previously been exported with a device, the profile will remain in any ZIPs of that Testrun in the future. This may cause confusion with the labs.

KrystsinaT (commented):
Screenshot from 2024-07-08 14-32-02
The error occurs if the profile was deleted and the report is downloaded without the risk profile.

jhughesbiot:
Updated zip method to create temp directory for test results and profile.  No need to move profiles to test directory and cleanup with this approach.

jhughesbiot (approved):

PR #580, opened 2024-07-05
Remove skipped result

jboddey:
Remove the skipped result from the base test module. Skipped results should no longer exist and have now been replaced with Feature Not Detected or Error.


KrystsinaT (commented):
@jboddey I see the Error result instead of skipped. Comfirm please that the result is right because I thought the result Feature not detected should be shown.

jhughesbiot (approved):

jboddey:
> @jboddey I see the Error result instead of skipped. Comfirm please that the result is right because I thought the result Feature not detected should be shown.

The skipped result is mapped to error or feature not detected - depending on the test

PR #581, opened 2024-07-05
Prevent creating device with duplicate manufacturer and model

jboddey:
Users should not be able to have duplicate devices in their device repository. This includes checking for a duplicate MAC address, or duplicate manufacturer and model combination. This pull requests adds an additional check to ensure a device with the same manufacturer and model does not already exist - returning error 409 if one is found.

### DIFF
--- a/framework/python/src/api/api.py
+++ b/framework/python/src/api/api.py
@@ -449,7 +449,16 @@ class Api:
         response.status_code = status.HTTP_400_BAD_REQUEST
         return self._generate_msg(False, "Invalid request received")
 
+      # Check if device with same MAC exists
       device = self._session.get_device(device_json.get(DEVICE_MAC_ADDR_KEY))
+
+      if device is None:
+
+        # Check if device with same manufacturer and model exists
+        device = self._session.get_device_by_make_and_model(
+          device_json.get(DEVICE_MANUFACTURER_KEY),
+          device_json.get(DEVICE_MODEL_KEY)
+        )
 
       if device is None:
 
@@ -468,7 +477,7 @@ class Api:
 
         response.status_code = status.HTTP_409_CONFLICT
         return self._generate_msg(
-            False, "A device with that " + "MAC address already exists")
+            False, "A device with that MAC address or name already exists")
 
       return device.to_config_json()
 
--- a/framework/python/src/api/api.py
+++ b/framework/python/src/api/api.py
@@ -452,7 +452,7 @@ class Api:
       # Check if device with same MAC exists
       device = self._session.get_device(device_json.get(DEVICE_MAC_ADDR_KEY))
 
-      if device is None:
+      if device is not None:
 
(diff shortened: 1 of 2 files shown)


KrystsinaT (commented):
Screenshot from 2024-07-08 15-33-53
The message is shown updated. After the user changes the name the save action doesn't function. The popup is not closed.

jboddey:
@KrystsinaT I believe this is a fault with the user interface. The request is not being sent at all.

jhughesbiot:
The UI already does a check for duplicate MAC address so while we probably won't ever get a request add a device with the same MAC, we should handle the messaging to not be a guess at what is duplicate.

image

```
      # Check if device with same MAC exists
      device = self._session.get_device(device_json.get(DEVICE_MAC_ADDR_KEY))
      dev_mac_exists = True if device is not None else False
      if device is None:

        # Check if device with same manufacturer and model exists
        device = self._session.get_device_by_make_and_model(
          device_json.get(DEVICE_MANUFACTURER_KEY),
          device_json.get(DEVICE_MODEL_KEY)
        )

      if device is None:
        # Create new device
        device = Device()
        device.mac_addr = device_json.get(DEVICE_MAC_ADDR_KEY).lower()
        device.manufacturer = device_json.get(DEVICE_MANUFACTURER_KEY)
        device.model = device_json.get(DEVICE_MODEL_KEY)
        device.device_folder = device.manufacturer + " " + device.model
        device.test_modules = device_json.get(DEVICE_TEST_MODULES_KEY)

        self._test_run.create_device(device)
        response.status_code = status.HTTP_201_CREATED

      else:
        message = 'A device with that '
        message += 'MAC address and' if dev_mac_exists else ''
        message += 'name already exists'
        response.status_code = status.HTTP_409_CONFLICT
        return self._generate_msg(
            False, message)
```
image


I am unclear however on the resolution to this issue as reported. The issue reports we should allow duplicate make and models with unique MAC addresses but this PR just reports the error better and doesn't allow it.

jboddey:
@jhughesbiot The intention is to not allow users to create a device with the same manufacturer and model.

jhughesbiot:
Can we update the issue description for clarity?  Seems to indicate the opposite currently.

> Expected: only the MAC address should be unique. Other fields are allowed to accept the same values as in another device

Otherwise, I think this is good to go.

framework/python/src/api/api.py
```diff
@@ -449,8 +449,21 @@ async def save_device(self, request: Request, response: Response):
         response.status_code = status.HTTP_400_BAD_REQUEST
         return self._generate_msg(False, "Invalid request received")
 
+      # Check if device with same MAC exists
       device = self._session.get_device(device_json.get(DEVICE_MAC_ADDR_KEY))
 
+      if device is None:
```
jhughesbiot, line 455: Should be

`if device is not None:`

PR #590, opened 2024-07-09
Fix GAR bug with cert upload

jboddey:
When the user uploads a cert and an error occurs, the user is informed that a cert with the same name already exists. However, this is not the case, and instead the cert was not the correct format - raising an exception. The logic behind determining the error message for the user was incorrect. This PR aims to fix that by correcting error logic.

### DIFF
--- a/framework/python/src/api/api.py
+++ b/framework/python/src/api/api.py
@@ -792,7 +792,7 @@ class Api:
       response.status_code = status.HTTP_409_CONFLICT
       return self._generate_msg(False,
                                 "A certificate with that name already exists.")
-    except IOError:
+    except Exception:
       LOGGER.error("An error occurred whilst uploading the certificate")
 
     # Return error if something went wrong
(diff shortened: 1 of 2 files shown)


jhughesbiot:
What were the steps to reproduce before this correction?  Removing the specific method to raise the Value error from the session introducess a bug to detect duplicate certificates as it now reports the wrong error:

image

Adding ValueError back in corrects this error:
image

Are we sure the reported issue was using a bad certificate format and not a duplicate?

jboddey:
The issue was that duplicate value errors were being thrown so the session was catching it's own error. I've updated the logic and tested the 3 scenarios: filename duplicate, common name duplicate and invalid certificate.

KrystsinaT (commented):
I have checked the 3 scenarios provided: filename duplicate, common name duplicate and invalid certificate. Everything works as expected now. Thank you!

PR #608, opened 2024-07-16
Allow ICMP response to DHCP messages in DHCP snooping test

jboddey:
In the event that a device has received a DHCP address from server 1, and still maintains communications with server 1, the device is allowed to reject messages from server 2. However, the DHCP snooping test checks the monitor.pcap for any packets not allowed. Currently, that test will report fail if the device has responded with an ICMP port unreachable to a DHCP ACK from server 2. Because the reject ICMP message includes the top part of the packet (which contains the DHCP ACK) this is picked up as a failure.

image


jhughesbiot:
What issue is being addressed here?  I'm not sure what this solves or if ICMP and DHCP can both be defined within the same packet.

jhughesbiot:
The screen shot still indicates DHCP and ICMP as two separate packets and the code is expecting both protocols to be contained in the same packet, which I don't believe is possible.

Can you add a unit test around this specific failure?

jhughesbiot (approved):

PR #611, opened 2024-07-16
Update all unit tests to work within the runtime environment

jhughesbiot:
Unit tests were not guaranteed to run within the same environment as they would in runtime so this updates the unit test structure to use containers as much as possible as well as the venv runtime for all non containerized tests.

- Update all test module unit tests to run from within the built container 
 - Update all host level unit tests to run within the venv 
 - Move some report tests out of test containers
 - Misc unit test cleanup


jboddey:
I cannot seem to run these unit tests. This is the error I get:

/usr/bin/python3: can't find '__main__' module in '/testing/unit/conn/module_test.py'
/usr/bin/python3: can't find '__main__' module in '/testing/unit/dns/module_test.py'
/usr/bin/python3: can't find '__main__' module in '/testing/unit/ntp/module_test.py'
/usr/bin/python3: can't find '__main__' module in '/testing/unit/protocol/module_test.py'
/usr/bin/python3: can't find '__main__' module in '/testing/unit/services/module_test.py'
^Cdocker: Error response from daemon: OCI runtime start failed: cannot start a container that has stopped: unknown.
testing/unit/run.sh: line 64: venv/bin/activate: No such file or directory

Could you quickly document in testing/unit/README.md how to run the unit tests?

jhughesbiot:
I'll add some docs around it's usage.  Were you running the unit test from the testing/unit directory?  That's currently required so might be a pathing issue if it was run from somewhere else.

jboddey:
I'm still getting some errors when running from the correct path: https://paste.googleplex.com/6099375204073472
Also, could we modify the run script to default from running from the root directory?

jboddey (approved):

PR #621, opened 2024-07-22
Build UI during package instead of install

jboddey:
This PR moves the build step of the UI from Testrun install to package. This will speed up the install time for users. An additional step has been added to the automated actions to test installing from package.


framework/python/src/core/testrun.py
```diff
@@ -463,7 +463,7 @@ def start_ui(self):
 
     try:
       client.containers.run(
-            image='test-run/ui',
+            image='testrun/ui',
             auto_remove=True,
```
jhughesbiot, line 467: Why change this format?  All other images use test-trun as naming convention.

jboddey: Rolling the change out slowly. Eventually they will all be changed to testrun as that is the correct naming.

sofyakurilova (approved):

cmd/build
```diff
@@ -36,9 +36,13 @@ fi
...
-# Build user interface
-echo Building user interface
-if docker build -t test-run/ui -f modules/ui/ui.Dockerfile . ; then
+# Check if UI has already been built (not if -l was used during install)
+if [ ! -d "modules/ui/dist" ]; then
+  cmd/build_ui
+fi
+
+# Build UI image
+if docker build -t testrun/ui -f modules/ui/ui.Dockerfile . ; then
```
jhughesbiot, line 45: When not running as root, this appears to block package from working as it has permissions errors on just about every single existing file in the installation directory. Why don't we want this run as root?

jboddey: Found an issue with this PR, looking to change it now.

jboddey: Everything in this PR should now be working. The only thing that you will have to do is clear out the make/usr and make/bin directory in your dev environment. I have also tested installing the new package over an old install and it works.

jhughesbiot: Running package stand alone since it's not inside a controlled environment has no dependency guarantee.  i.e. build_ui needs npm but npm is only installed during the install script, which is not linked to the package method.

jboddey: The idea is that if you are installing via package, then whoever packages testrun will have npm installed (usually myself). If the user is installing from source using cmd/install -l, they should have run cmd/prepare before (which will install npm).

jhughesbiot: Should we add a call to prepare into the package script to make sure the system is always ready?

jhughesbiot: Should automate the removal of existing make directories as well to make this process cleaner.

jhughesbiot: Can't seem to get nmp to install on my system due to various dependency issues.  Should likely move the entire build into a docker container to maintain stability in the build and extract the generated dist from the container.

jboddey: I will move the build process into a container to remove the dependency.

cmd/build_ui
```diff
@@ -0,0 +1,35 @@
...
+else
+  echo An error occured whilst building the ui builder
+  exit 1
+fi
+
+# Check that the container is not already running
+docker kill tr-ui-build 2> /dev/null || true
+
+# Start build container and build the ui dist 
+docker run -it --rm -v $PWD/modules/ui:/modules/ui testrun/build-ui /bin/sh -c "npm install && npm run build"
```
jboddey, line 32: Use of -it is causing the error message on actions.

jhughesbiot (approved):

PR #625, opened 2024-07-22
Fix weasyprint/pydyf  dependency mismatch that causes an API runtime error (too many arguments passed to pydif function by weasyprint)

caseykelso:
When building testrun in the "wild", our computers hit PYPI directly, I'm assuming that Google is probably using a pypi cache internally so the behavior is different.  weasyprint pulls down the latest version of pydyf (0.11.0) which is incompatible with weasyprint 61.2 as specified in testrun's requirements.txt.  

Two possible paths for a fix:
- specify pydyf==0.8.0 in testrun's requirements.txt
- increment testrun's requirements.txt weasyprint requirement from 61.2 to 62.3
- I decided to move weasyprint to 62.3, and add pydyf to the requirements.txt and lock it to 0.11.0
- We have tested it internally and it resolves the runtime issues with testrun.

### DIFF
--- a/framework/requirements.txt
+++ b/framework/requirements.txt
@@ -9,7 +9,6 @@ scapy==2.5.0
 
 # Requirments for the test_orc module
 weasyprint==62.3
-pydyf==0.11.0
 
 # Requirements for the API
 fastapi==0.109.1
--- a/framework/requirements.txt
+++ b/framework/requirements.txt
@@ -9,6 +9,7 @@ scapy==2.5.0
 
 # Requirments for the test_orc module
 weasyprint==62.3
+pydyf==0.11.0
 
 # Requirements for the API
 fastapi==0.109.1
--- a/framework/requirements.txt
+++ b/framework/requirements.txt
@@ -7,8 +7,11 @@ ipaddress==1.0.23
 netifaces==0.11.0
 scapy==2.5.0
 
-# Requirments for the test_orc module
+# Requirements for the test_orc module
 weasyprint==61.2
+
+# add more strict requirement for pydyf as a dependency of weasyprint, as newer pydyf has breaking API changes
+pydyf==0.8.0
 
 # Requirements for the API
 fastapi==0.109.1
--- a/framework/requirements.txt
+++ b/framework/requirements.txt
@@ -8,7 +8,7 @@ netifaces==0.11.0
 scapy==2.5.0
 
 # Requirments for the test_orc module
-weasyprint==61.2
+weasyprint==62.3
 
 # Requirements for the API
 fastapi==0.109.1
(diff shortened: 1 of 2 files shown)


jboddey:
Thanks for the pr @caseykelso. An issue that comes up in this PR is the version number for weasyprint. Despite pypi suggests the latest version is 62.3, it seems upto 61.2 is available. It may be due to python version used etc. Check the actions for the error message.

caseykelso:
> Thanks for the pr @caseykelso. An issue that comes up in this PR is the version number for weasyprint. Despite pypi suggests the latest version is 62.3, it seems upto 61.2 is available. It may be due to python version used etc. Check the actions for the error message.

---
Hi, does this mean you would prefer to keep the weasyprint at 61.2 in the requirements.txt, and just add a requirement for pydyf to 0.8.0 to resolve the issue?

jboddey:
> > Thanks for the pr @caseykelso. An issue that comes up in this PR is the version number for weasyprint. Despite pypi suggests the latest version is 62.3, it seems upto 61.2 is available. It may be due to python version used etc. Check the actions for the error message.
> 
> Hi, does this mean you would prefer to keep the weasyprint at 61.2 in the requirements.txt, and just add a requirement for pydyf to 0.8.0 to resolve the issue?

That looks to be the only way to get it working unless we can understand why version 62.3 does not seem to be available to us

jboddey:
@caseykelso I have just tried upgrading to 62.3 and it now allows me to. You may have been correct with regards to the Google cache. In version 62.3, the issue with pydyf is resolved so there should be no need to pin it. The issue with weasyprint can be found here: https://github.com/Kozea/WeasyPrint/issues/2200

Could you verify that this PR works for you?

jboddey:
@caseykelso Could you sign the Google CLA as soon as possible so we can get this released in a hotfix? You can do this here: https://cla.developers.google.com/

caseykelso:
> @caseykelso Could you sign the Google CLA as soon as possible so we can get this released in a hotfix? You can do this here: https://cla.developers.google.com/

No problem, I signed it.

caseykelso:
> @caseykelso I have just tried upgrading to 62.3 and it now allows me to. It seems that 62.3 relies on Python 3.9 which is not the default python version in ubuntu 20 so that is not a viable upgrade just yet. The issue with weasyprint can be found here: [Kozea/WeasyPrint#2200](https://github.com/Kozea/WeasyPrint/issues/2200)
> 
> Could you verify that this PR works for you?

We've already covered this PR in our testing yesterday. weasyprint 62.3 locks the pydyf version to 0.11.0 or greater, and less than 0.12.0, so we'll only get bug patches in theory.

https://github.com/Kozea/WeasyPrint/blob/main/pyproject.toml#L15

jboddey (approved):

PR #629, opened 2024-07-23
Fix individual test disabling when run from the UI

jhughesbiot:
When the UI sends the device test modules to the start command, it does not include the individual tests, only the test module so it overrides individual test configuration.  This PR merges these options properly when run from the UI.  It also handles the disabled state for the report.


jboddey (approved):

jboddey:
@jhughesbiot I have removed the disabled test result for now as this is not rendered correctly in the UI or the PDF. We can schedule to add this for version 1.3.3 (I have left it in the code but commented it out)

PR #636, opened 2024-07-25
Allows draft profiles to become expired

jboddey:


framework/python/src/common/session.py
```diff
@@ -528,16 +528,14 @@ def update_profile(self, profile_json):
...
+    today = datetime.datetime.now().timestamp()
 
-      today = datetime.datetime.now().timestamp()
-
-      if created_date < (today - SECONDS_IN_YEAR):
-        profile.status = 'Expired'
+    if created_date < (today - SECONDS_IN_YEAR):
+      profile.status = 'Expired'
 
     return profile.status
```
jhughesbiot, line 540: This should be a direct call to the profile where expiration status is already determined.  Since that is a private method, likely a new method should be created there for updating the status only,i.e. 
```

profile.update_status()
return profile.status
```

testing/unit/services/output/services.log
jhughesbiot, line 1: Shouldn't be included in commits

jboddey: As discussed, this profile should have not made it's way into git

framework/python/src/common/risk_profile.py
```diff
@@ -95,6 +96,17 @@ def get_file_path(self):
...
+    """Checks if the profile has expired
+    and updates the status to Expired if required."""
+
+    created_date = self.created.timestamp()
+
+    today = datetime.now().timestamp()
+
+    if created_date < (today - SECONDS_IN_YEAR):
+      self.status = 'Expired'
+
```
jhughesbiot, line 109: Use _expired method directly to avoid using seconds in year

```
def check_status(self):
    """Checks if the profile has expired
    and updates the status to Expired if required."""
    if self._expired():
      self.status = 'Expired'
```

jhughesbiot: Still using seconds in year instead of directly calling the expired method, which already handles this check.  Seconds in year is inaccurate as it does not account for leap years, should not be used.

If we are going the route of dynamically resolving the status, we should likely privatize the status property and switch all references to a get_status() method call which makes this determination every time.

framework/python/src/common/risk_profile.py
```diff
@@ -73,6 +73,11 @@ def load(self, profile_json, profile_format):
     self._validate(profile_json, profile_format)
     self.update_risk(profile_format)
 
+    # Check if the profile has expired
+    expired = self._expired()
+    if expired:
+      self.status = 'Expired'
+
```
jhughesbiot, line 80: The _validate method already updates the status correctly so if we want to maintain this within the load/constructor process, we should update the _validate method directly to account for this.

```
  def _validate(self, profile_json, profile_format):
    if self._expired():
      self.status = 'Expired'
    elif self._valid(profile_json, profile_format):
      # User only wants to save a draft
      if 'status' in profile_json and profile_json['status'] == 'Draft':
        self.status = 'Draft'
      else:
        self.status = 'Valid'
    else:
      self.status = 'Draft'
```

framework/python/src/common/risk_profile.py
```diff
@@ -24,6 +24,7 @@
 PROFILES_PATH = 'local/risk_profiles'
 LOGGER = logger.get_logger('risk_profile')
 RESOURCES_DIR = 'resources/report'
+SECONDS_IN_YEAR = 31536000
```
jhughesbiot, line 27: Delete unused variable

jhughesbiot (approved):

PR #641, opened 2024-07-29
Text changes

sofyakurilova:
Update texts


modules/ui/src/app/interceptors/error.interceptor.ts
```diff
@@ -57,17 +58,18 @@ export class ErrorInterceptor implements HttpInterceptor {
...
         } else {
           if (error.status === 0) {
             this.notificationService.notify(
-              'Back End is not responding. Please, try again later.'
+              'Testrun is not responding. Please try again in a moment.'
             );
           } else {
             this.notificationService.notify(
-              error.error?.error || error.message
+              `Something went wrong. Check the logs for details here ${LOGS_FOLDER}`
```
jboddey, line 70: I'm just querying with Aksinnia whether we include the full logs url as 1) it can vary if the user installs differently and 2) The logs should be viewed in the console windows for the current runtime.

hitnik (approved):

PR #642, opened 2024-07-29
Updating the testrun status via MQTT.

hitnik:


framework/python/src/common/session.py
```diff
@@ -38,13 +38,39 @@
 CERTS_PATH = 'local/root_certs'
 CONFIG_FILE_PATH = 'local/system.json'
 SECONDS_IN_YEAR = 31536000
+STATUS_TOPIC = 'status'
```
sofyakurilova, line 41: Should it be /events/status?

hitnik: According to notes from Bi-weekly sync, it was proposed to call this topic "status"

jboddey: Correct yes, as this is not necessarily an event, it should just be could /status.

framework/python/src/common/mqtt.py
```diff
@@ -43,6 +43,7 @@ def _connect(self):
     if not self._client.is_connected():
       try:
         self._client.connect(self._host, WEBSOCKETS_PORT, 60)
+        self._client.connect(self._host, WEBSOCKETS_PORT, 60)
```
jboddey, line 46: Should this line be duplicated?

framework/python/src/common/session.py
```diff
@@ -335,6 +364,27 @@ def add_test_result(self, result):
...
+      if test_result.name == result.name:
+
+        # Just update the result and description
+        if test_result.result == 'In Progress':
+          test_result.result = 'Error'
+          test_result.description = result.description
+          test_result.recommendations = None
+          updated = True
+
+    if not updated:
```
jboddey, line 383: Is there a reason why it would not be updated in the check above? Why not just use this part standalone?

jboddey:
@hitnik could you provide an example output of the JSON that is sent with this?

framework/python/src/common/mqtt.py
hitnik: removed

framework/python/src/common/session.py
hitnik: fixed

hitnik:
> @hitnik could you provide an example output of the JSON that is sent with this?

Same as the /system/status response.

jboddey:
@hitnik Just one pylint issue to resolve and I will try this out

framework/python/src/common/session.py
```diff
@@ -37,13 +37,39 @@
...
 
 
+def session_tracker(method):
+  """Session changes tracker."""
+  def wrapper(self, *args, **kwargs):
+
+    result = method(self, *args, **kwargs)
+
+    if self.get_status() != 'Idle':
+      self.get_mqtt_client().send_message(STATUS_TOPIC, self.to_json())
```
jboddey, line 55: The JSON format is not correct. fastapi automatically converts some Python objects into JSON for us. Now that it is not used, raw Python objects are being included in the response. See here: https://paste.googleplex.com/6069255923826688 Before this can be merged, more work is required to verify the JSON being received by the UI.

jboddey (changes requested):

framework/python/src/common/session.py
hitnik: I fixed it in the ui branch

jboddey:
Looks good now. However, we will merge this is once v1.4 is released

PR #643, opened 2024-07-29
Update the requests dependency

jboddey:


jboddey:
Upgrading requests to 2.32.3 breaks the docker-py module as seen here: https://github.com/docker/docker-py/pull/3257 Going to try upgrading the docker module too.

jhughesbiot (approved):

PR #646, opened 2024-07-31
Adds internet icon

sofyakurilova:
Video after changes
http://screencast/cast/NTc5MDA1MzgxMTg3OTkzNnw3NmQ2ZGMxMS01MA

Tests
ABvLwPUTYgApzr2


jboddey:
Could we remove the popup at the bottom of the screen? When Testrun starts up it will report false for internet because of changing network settings so will be a false negative. I think keeping it simple with just the icon is fine.

jboddey (approved):

PR #653, opened 2024-08-01
Show internet connection

hitnik:


framework/python/src/net_orc/ip_control.py
```diff
@@ -241,6 +241,24 @@ def configure_container_interface(self,
...
 
+  def ping_via_gateway(self, host):
+    """Ping the host trough the gateway container"""
+    command = f'docker exec tr-ct-gateway ping -W 1 -c 1 {host}'
+    output = util.run_command(command)
+    if '0% packet loss' in output[0]:
+      return True
+    return False
+  
+  def ping_via_interface(self, host, iface=None):
```
jboddey, line 252: Where is this method being used?

hitnik: removed

PR #658, opened 2024-08-04
Add device types and technologies endpoints

jboddey:


jhughesbiot:
Seems like this is pretty out of sync with the expected merge location. Can we get the merge strategy cleaned up for this PR so it can be reviewed properly?

jboddey:
> Seems like this is pretty out of sync with the expected merge location. Can we get the merge strategy cleaned up for this PR so it can be reviewed properly?

This is resolved now.

jboddey:
@hitnik could you take a look at this PR?

jhughesbiot (approved):

PR #668, opened 2024-08-08
Endpoint for a device profile questions

hitnik:
I refactored the code so that a single endpoint will be used for all questions and answers. The questions are divided into steps, and if there is a subsequent step, a link to the next step will be included in the response. Each question and each answer has been assigned an ID. We discussed this solution with the team and concluded that it would be more convenient this way. Of course, at the moment, there is no need for such a structure, but it is implemented with the addition of new questions in mind.

### DIFF
--- a/resources/devices/device_profile.json
+++ b/resources/devices/device_profile.json
@@ -1,0 +1,353 @@
+[
+  {
+    "step": 1,
+    "questions": [
+      {
+        "id": 1,
+        "question": "What type of device is this?",
+        "validation": {
+          "required": true
+        },
+        "type": "select",
+        "options": [
+          {
+            "text": "Building Automation Gateway",
+            "risk": "High",
+            "id": 1
+          },
+          {
+            "text": "IoT Gateway",
+            "risk": "High",
+            "id": 2
+          },
+          {
+            "text": "Controller - AHU",
+            "risk": "High",
+            "id": 3
+          },
+          {
+            "text": "Controller - Boiler",
+            "risk": "High",
+            "id": 4
+          },
+          {
+            "text": "Controller - Chiller",
+            "risk": "High",
+            "id": 5
+          },
+          {
+            "text": "Controller - FCU",
+            "risk": "Limited",
+            "id": 6
+          },
+          {
+            "text": "Controller - Pump",
+            "risk": "Limited",
+            "id": 7
+          },
+          {
+            "text": "Controller - CRAC",
+            "risk": "High",
+            "id": 8
+          },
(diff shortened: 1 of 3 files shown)


testing/api/test_api.py
```diff
@@ -1444,3 +1446,26 @@ def test_delete_profile(testrun, reset_profiles, add_profile): # pylint: disable
...
   # Check if profile was deleted
   assert deleted_profile is None
+
+
+def test_get_device_profile():
+  """ Test get device profile question"""
+  with open(DEVICE_PROFILE_QUESTIONS, "r", encoding="utf-8") as file:
+    questions = json.load(file)
+
+  # Checking the 400 error when a step does not exist.
```
jboddey, line 1456: Should a 404 be returned for this instead of bad request if the steps are not found?

resources/devices/device_profile.json
```diff
@@ -0,0 +1,353 @@
...
+      {
+        "id": 3,
+        "question": "Please select the technology this device falls into",
+        "validation": {
+          "required": true
+        },
+        "type": "select",
+        "options": [
+          {
+            "id": 1,
```
jboddey, line 254: Previously, for strings and select answers, the string has been passed on its own so I'm not sure of the need for an ID here

jboddey (commented):
I'm unsure of the use of an ID in these questions. For single select answers we have previously just used the string

testing/api/test_api.py
hitnik: > Should a 404 be returned for this instead of bad request if the steps are not found?

Done

hitnik:
> I'm unsure of the use of an ID in these questions. For single select answers we have previously just used the string

@jboddey  I primarily did this for multiselect answer options, but I added IDs to other questions for consistency.

resources/devices/device_profile.json
```diff
@@ -0,0 +1,353 @@
+[
+  {
+    "step": 1,
```
sofyakurilova, line 3: Can you please add fileds "title" and "description" on this level? 
Content: 
"title": "Select device type & technology", 
"description": "Before your device can go through testing, tell us more about your device and its functionality. It is important that we fully understand your device before a thorough assessment can be made."

hitnik: > "title": "Select device type & technology",
> "description": "Before your device can go through testing, tell us more about your device and its functionality. It is important that we fully understand your device before a thorough assessment can be made."

Done

jboddey (approved):

PR #669, opened 2024-08-08
Fix a delay in the internet connectivity check

jboddey:


hitnik:
Perhaps we should add a timeout parameter to util.run_command, because the Popen.communicate method used under the hood already has a timeout. It would be more transparent and useful for the future.

jboddey:
@hitnik I've just tried that and would agree. However, for some reason there is still a delay when using the timeout argument of subprocess.communicate

hitnik:
> @hitnik I've just tried that and would agree. However, for some reason there is still a delay when using the timeout argument of subprocess.communicate

"Hmm, this is interesting, I'll take a look at it.

jboddey:
@hitnik Any luck with this?

hitnik:
> @hitnik Any luck with this?

Yes, I will push my changes later.

KrystsinaT (commented):
I have tested. The internet connection view on UI  changes in ~3-5 sec after the phisical disconnection/connection was performed.

hitnik:
I made some optimizations: First, I installed the fping  to  the gateway container, which works several times faster than the standard ping. Second, I set the job to run only when the single-intf mode is disabled. Third, I send messages about the internet connection to the frontend only when the testing process is running. I also changed the check interval to 2 seconds and set the timeout to 1 second.

jboddey:
@hitnik, @KrystsinaT has tested out the latest commit and it seems the disconnection time has reverted back to 15-20 seconds. Could you revert your commit? We will have to proceed with the timeout command.

hitnik:
@jboddey  I resolved the delay issue when the internet connection is physically disconnected.

KrystsinaT (commented):
In case of the cable disconnected from adapter the internet connection icon changes after more than 20 sec. In other cases ~3-5 seconds.

hitnik:
@KrystsinaT  I reverted to  Jacob's solution.

KrystsinaT (commented):
Retested. Works quickly - 4-5 seconds. Thank you! @jboddey @hitnik

jboddey:
@hitnik Could you approve this PR if you are happy with the changes please?

hitnik (approved):

PR #672, opened 2024-08-08
modularizing docker methods

jhughesbiot:
- Add python representations of docker modules
- Update testrun to use new modules
- Update session to contain additional shared variables to work with new modules
- Update test module images to "testrun" prefix


framework/python/src/common/session.py
```diff
@@ -100,6 +100,13 @@ def __init__(self, root_dir):
     self._load_config()
     self._load_profiles()
 
+    # Network information
+    self._ipv4_subnet = None
```
jboddey, line 104: Can we take networking info and put it into another core class? I don't think the session is the right place for it

jhughesbiot: Since session is passed around to multiple services, I feel this might actually be the best place for all configurations to load into.  We also load the session first so making this the location that initiates all configurations first and then those services utilize those configurations seems the most efficient way to handle these things.

jboddey: Ok, to reduce the size of the session, could we not have a separate class like NetworkConfig or NetworkSession that is created and stored in the session?

jhughesbiot: I think we could do something like that but make this a separate PR to work on the session alone since this is already a pretty large PR.

jhughesbiot:
@jboddey Network modules are now added and should be ready for a full review

jboddey:
All looks good to me. Once dev is merged into main for v1.4 we can go ahead and merge this.

jboddey (approved):

PR #678, opened 2024-08-13
Check if device folder already exists

jboddey:


KrystsinaT (commented):
Tested. 409 status returns. The snack bar with the proper message is shown. Thank you! @jboddey

hitnik (approved):

PR #709, opened 2024-08-22
Passing "single interface" mode to the FE

hitnik:


jboddey:
@hitnik What's the use case for this?

hitnik:
> @hitnik What's the use case for this?

@jboddey  This is done so that the FE is aware of the single interface mode and does not change the internet connection icon, disables the internet interface selection field, and implements a few other related features.

jboddey (approved):

PR #723, opened 2024-08-27
Adding concept of test packs

jboddey:
This PR adds two new test packs (pilot & qualification). These json files list the tests which are required and not. The framework loads in the test packs and overrides the required result of tests in the test_orc.


framework/python/src/test_orc/test_orchestrator.py
```diff
@@ -148,6 +174,19 @@ def run_test_modules(self):
...
+    # Set testing description
+    test_pack: TestPack = self.get_test_pack(device.test_pack)
+
+    # Default message is empty (better than an error message).
+    # This should never be shown
+    message: str = ""
+    if report.get_status() == TestrunStatus.COMPLIANT:
+      message = test_pack.get_message("compliant_description")
+    elif report.get_status() == TestrunStatus.NON_COMPLIANT:
+      message = test_pack.get_message("non_compliant_description")
```
jhughesbiot, line 186: What is the purpose of these messages?  I don't seem to see their usage anywhere.

jboddey: Those messages are going to be used on the new Testrun completion modal, I will send the design over

jhughesbiot (approved):

PR #729, opened 2024-08-29
Update delete device endpoint 

MariusBaldovin:
Updated the delete device endpoint to respond 404 instead of 500 if the report is not found.

Updated two tests for delete device endpoint to test 404 response (test_delete_report_no_report and test_delete_report_no_device).


testing/api/devices/device_2.json
```diff
@@ -0,0 +1,56 @@
...
+        },
+        {
+          "question": "Is there a second IP port on the device?",
+          "answer": "Yes"
+        },
+        {
+          "question": "Can the second IP port on your device be disabled?",
+          "answer": "Yes"
+        }
+        ],
```
jboddey, line 34: Can you fix this formatting?

jboddey (changes requested):

testing/api/devices/device_2.json
MariusBaldovin: Solved

testing/api/devices/device_1.json
```diff
@@ -0,0 +1,55 @@
+{
+  "status": "Valid",
```
jboddey, line 2: The status field should not be passed into the API. Only the backend will determine whether it is valid or not.

testing/api/devices/device_2.json
```diff
@@ -0,0 +1,55 @@
+{
+  "status": "Valid",
```
jboddey, line 2: Same with this one

jboddey (changes requested):

jboddey (approved):

PR #739, opened 2024-09-02
362724933: (feat) add program type to results tab

OlgaMardvilko:
### Overview

Ticket: 362724933
feat: add program type to results tab

### Screenshot after changes
https://screenshot.googleplex.com/57H7WxdGQUEcTiu


jboddey:
Would it not be more space effective if we prepend the program icon to the device name?

OlgaMardvilko:
> Would it not be more space effective if we prepend the program icon to the device name?

For space definitely yes, but we have the requirements to display them as text. So, if we want to change it, let discus with @Aksinnia-Dzenisenka

sofyakurilova (approved):

PR #747, opened 2024-09-04
Generate PDF again when ZIP export called

jboddey:
This PR enables generation of the PDF report each time the user attempts to export the ZIP file. This has been done so that we can test PDF changes more easily. This PR also loads in the module reports for old reports each time Testrun is restarted.


hitnik (approved):

jboddey:
This is no longer required.

PR #764, opened 2024-09-09
Pin python requirements

jhughesbiot:
Use semi-automatic process of determining existing stable python dependencies and manually updating requirements.txt file.


jboddey:
Looks good. However, could you briefly explain how this works and the use of these files:
image

Also, is there anyway we could run this command without sudo?

swapnilGirishPawar (commented):
New changes looks good to me,

jboddey:
Note: Comment out bin script to use as reference, rest is fine to merge.

jhughesbiot:
> Note: Comment out bin script to use as reference, rest is fine to merge.

This should be ready to go now.

jboddey (approved):

PR #783, opened 2024-09-17
Feature/tls 1.0

jhughesbiot:
- Adds TLS 1.0 client test
- Defines this as informational in qualification profile.
- TLS 1.0 or 1.1 client detection will still fail TLS 1.2 and 1.3 client tests

Qualification profile output:
image

Pilot profile output:
image


modules/test/tls/python/src/tls_module.py
```diff
@@ -265,18 +266,60 @@ def _security_tls_v1_3_server(self):
...
+        result_state = False
+      LOGGER.info(f'Result State: {result_state}')
+      if result_state:
+        result_message = 'TLS 1.0 or higher detected'
+      else:
+        result_message = 'TLS 1.0 or higher was not detected'
+      result_details = tls_1_0_valid[2] + tls_1_1_valid[2] + tls_1_2_valid[
+          2] + tls_1_3_valid[2]
+      result_tags = []
+      result_tags.extend(tls_1_0_valid[3])
```
jboddey, line 306: Is this adding multiple tags to the result?

jhughesbiot: Good catch, updated to prevent duplicate entries.

jboddey:
I've received the following results:
image

The v1.0 client test should return Feature Not Detected as there were no outbound TLS connections.

jhughesbiot:
Fixed Feature not detected result

image

jboddey (approved):

PR #795, opened 2024-09-20
Move single_intf from docker environment variables to /system/config/ eendpoint

hitnik:


jboddey:
What's the reason for moving this? In my opinion single interface should not be a configuration option as it may be changed frequently

hitnik:
> What's the reason for moving this? In my opinion single interface should not be a configuration option as it may be changed frequently

@jboddey 
Olga and I tried to pass this parameter as a Docker container environment variable, but nothing can be passed to the frontend after the code is built. We identified two options: either create a separate API endpoint or add it to the sys config, as this parameter logically belongs to the system configuration. Do you think creating a separate endpoint would be better?

jboddey:
@hitnik Is the reason for wanting to pass single_intf to the UI to disable the internet icon? I think a better method for this would be to pass 'None' instead of true/false in the MQTT for internet_connection if we can't get environment variables working.

hitnik:
> @hitnik Is the reason for wanting to pass single_intf to the UI to disable the internet icon? I think a better method for this would be to pass 'None' instead of true/false in the MQTT for internet_connection if we can't get environment variables working.



> @hitnik Is the reason for wanting to pass single_intf to the UI to disable the internet icon? I think a better method for this would be to pass 'None' instead of true/false in the MQTT for internet_connection if we can't get environment variables working.

On the backend, sending internet connection information through MQTT is disabled in single interface mode.

jboddey (approved):

OlgaMardvilko (approved):

PR #827, opened 2024-09-30
Firmware is not entered in testrun dialog after navigation to it with screen reader ON

sofyakurilova:
Add delay and fix dialog open configuration


OlgaMardvilko (approved):

OlgaMardvilko (changes requested):
Please fix tests

PR #848, opened 2024-10-02
Fix report_test.py tests

MariusBaldovin:
Fixed unit tests for reports. Updated add_html_formatting() from report_test.py  to format the reports for modules using Jinja2 library.


testing/unit/report/report_test.py
```diff
@@ -95,34 +140,57 @@ def report_formatting(self):
...
+    # Load the css file
+    with open(CSS_PATH, 'r', encoding='UTF-8') as css_file:
+      styles = css_file.read()
+
+    # Define the html template
+    html_template = '''
     <!DOCTYPE html>
     <html lang="en">
-    {TestReport().generate_head()}
+    <head>
```
jboddey, line 154: Can you put this header into a html file and load it in that way?

MariusBaldovin: Updated to use regex to extract the head from 'test_report_template.html' file

jhughesbiot:
The module reports don't seem to be generating the body.

MariusBaldovin:
> The module reports don't seem to be generating the body.

Removed from the body double curly braces and now the body content is generated in html files.

PR #854, opened 2024-10-04
Verify all containers stopped before cancel testrun

hitnik:


jboddey:
What's the reason behind this?

jboddey:
Looking at the log in the bug, the changes in this PR would not fix it. The issue in the bug is caused because the network listener is still running, triggering the _device_discovered callback in the net_orc.

PR #861, opened 2024-10-07
Certificates endpoint

MariusBaldovin:
Updated '/system/config/certs':
- removed 'add_cert' fixture which was using the API and added 2 new fixtures 'add_one_cert' and 'add_two_certs' to add the certs to local folder. 
- updated the tests to use the new fixtures


jboddey:
@MariusBaldovin could you not replace the two fixtures with one that can take an array of certificate file names?

MariusBaldovin:
> @MariusBaldovin could you not replace the two fixtures with one that can take an array of certificate file names?

@jboddey becasue The fixtures are copying the cert files and the fixtures are running before testrun fixture for the certificates to be loaded. If I create one fixture that can take the certificate file names and I call the fixture inside the test this will mean that this new fixture will run after testrun fixture and the certificates will not be loaded. 

A different approach will be using @pytest.mark.parametrize that might work and this way only one fixture will be used.

MariusBaldovin:
@jboddey The tests for certs endpoint are now using add_certs fixture. I have also concatenated the 3 tests for no certs, one cert and two certs available into one test and I am using @pytest.mark.parametrize to create the 3 scenarios.

jhughesbiot (approved):

PR #871, opened 2024-10-09
Feature/tls module report

jhughesbiot:
- Add TLS module report back with updated module report style
- Increase TLS module timeout


jboddey:
Note: @jboddey to check the design of this.

jboddey:
New PDF report design:
image

jhughesbiot:
> Note: I have removed the issuer information as I don't think that is really required here. We are only interested in the device cert and that it has been signed by a known CA.

The Issuer gives the signer information though, isn't that relevant?

Also, do you have a device that resolves multiple certificates to test?  That is where I have seen formatting issues on pagination.

jboddey:
I think as long as the cert has been signed, the signed by field will be provided and the user should know the details of the issuer. Also, I'm struggling to find a place for the issuer information. I will try with a device with multiple TLS certificates.

duncangreene:
Will this proposed section in the report provide context as to _why_ the the TLS 1.2 and TLS 1.3 certs are being reported as invalid when the user sees them in the results list below?

image

Sort of akin to how the user might see it in Google Chrome?

image

duncangreene:
> This PR will not address that particular request. Please could you raise a feature request detailing this and we will take a look?

#955

hitnik:
@jboddey  I have added splitting by module summary and refactored the formatting of the report module.   Could you check this on a real report?

jboddey:
@hitnik Looks a lot better. Just some small things:

image

We could do with right padding on the signed by property.
We need margin top on the summary bar for the non-first pages.

hitnik:
@jboddey  I reduced the module-summary font from 24 to 22, added padding on the right, and also added a margin-top of 10 for non-first pages.

jboddey (approved):

PR #896, opened 2024-10-17
374084009: Fix error on restoring DHCP server configuration if device has no DHCP lease

jboddey:
This PR fixes:
 - A bug in private address test in the wait for lease to expire method where the lease is None
 - Removes file logs for modules as they were no longer used (module.log is used now instead)
 - Cleaned up the dhcp log file
 - Adds an extra step to resolve for dhcp address test
 - Improve the readability of some test descriptions


modules/network/dhcp-1/python/src/grpc_server/dhcp_leases.py
```diff
@@ -56,20 +56,21 @@ def get_lease(self, hw_addr):
...
+
       try:
         lease = DHCPLease(line)
         leases.append(lease)
-      except Exception as e:  # pylint: disable=W0718
+      except Exception:  # pylint: disable=W0718
         # Let non lease lines file without extra checks
-        LOGGER.error('Making Lease Error: ' + str(e))
-        LOGGER.error('Not a valid lease line: ' + line)
+        LOGGER.info('Not a valid lease line: ' + line)
```
jhughesbiot, line 73: Should maintain actual Exception message reporting as this doesn't give any information to what the actual underlying error was without it.

modules/network/dhcp-2/python/src/grpc_server/dhcp_leases.py
```diff
@@ -56,20 +56,21 @@ def get_lease(self, hw_addr):
...
+
       try:
         lease = DHCPLease(line)
         leases.append(lease)
-      except Exception as e:  # pylint: disable=W0718
+      except Exception:  # pylint: disable=W0718
         # Let non lease lines file without extra checks
-        LOGGER.error('Making Lease Error: ' + str(e))
-        LOGGER.error('Not a valid lease line: ' + line)
+        LOGGER.info('Not a valid lease line: ' + line)
```
jhughesbiot, line 73: Same as previous note on dhcp1 leases

modules/network/dhcp-1/python/src/grpc_server/dhcp_leases.py
```diff
@@ -56,20 +56,21 @@ def get_lease(self, hw_addr):
...
     lines = lease_list.split('\n')
     for line in lines:
+
+      # Ignore non-lease lines
+      if line.contains('Reading leases from') or line.contains('==='):
+        continue
+
       try:
         lease = DHCPLease(line)
         leases.append(lease)
```
jhughesbiot, line 70: This change appears to break lease resolving and all DHCP tests, needs to be corrected or reverted.

jhughesbiot (approved):

PR #916, opened 2024-10-23
Feature/cli no UI

jhughesbiot:
Add target device options to CLI when --no-ui option is  used:

Error checking for following conditions (Errors do not apply when using --net-only option):

1. No target device mac address or firmware provided when --no-ui is used:

```
sudo bin/testrun --no-ui
usage: test_runner.py [-h] [-f CONFIG_FILE] [--validate] [-net]
                      [--single-intf] [--no-ui] [--target TARGET]
                      [-fw FIRMWARE]
test_runner.py: error: --target and --firmware required when --no-ui is specified
```

2. Target device provided does not exist:
```
sudo bin/testrun --no-ui --target 68:5e:1c:cb:6e:cb -fw 1.0


Traceback (most recent call last):
  File "/home/ubuntu/google/test-run/framework/python/src/core/test_runner.py", line 117, in <module>
    runner = TestRunner(config_file=args.config_file,
  File "/home/ubuntu/google/test-run/framework/python/src/core/test_runner.py", line 43, in __init__
    self._testrun = Testrun(config_file=config_file,
  File "/home/ubuntu/google/test-run/framework/python/src/core/testrun.py", line 122, in __init__
    raise ValueError(
ValueError: Target device specified does not exist in device registry
```


jboddey:
The usage statement looks a bit confusing: ```usage: test_runner.py```. Could we make this more accurate, e.g ```usage: testrun...```

Also, the stack trace does not look very friendly when a device can not be found, could we not have the device checked in Python and then exit out?

jhughesbiot:
> The usage statement looks a bit confusing: usage: test_runner.py. Could we make this more accurate, e.g usage: testrun...

test_runner is where all the args are resolved and checked so changing the name I think would cause more confusion.  This is the fastest point of exit unless I move the logic to check arguments to testrun instead of the args direrctly.

> Also, the stack trace does not look very friendly when a device can not be found, could we not have the device checked in Python and then exit out?

The check is in python (testrun.py) and raises the error but the caller (test_runner) does not have any try/except in it's code to suppress the stacktrace.  Stack trace inclusion at this level seems more important since uncaught failures that cause critical failures without a stack won't be easily debugged.

jhughesbiot:
@jboddey 

1. Updated bad arg messaging

```
sudo bin/testrun --no-ui

Testrun

optional arguments:
  -h, --help            show this help message and exit
  -f CONFIG_FILE, --config-file CONFIG_FILE
                        Define the configuration file for Testrun and Network
                        Orchestrator (default: None)
  --validate            Turn on the validation of the network after network
                        boot (default: False)
  -net, --net-only      Run the network only, do not run tests (default:
                        False)
  --single-intf         Single interface mode (experimental) (default: False)
  --no-ui               Do not launch the user interface (default: False)
  --target TARGET       MAC address of the target device (default: None)
  -fw FIRMWARE, --firmware FIRMWARE
                        Firmware version to be tested (default: None)
Error: --target and --firmware are required when --no-ui is specified
```


2. Updated bad target messaging

```
sudo bin/testrun --no-ui --target aa:bb:cc:dd:ee:ff -fw 1.0

Target device specified does not exist in device registry: aa:bb:cc:dd:ee:ff
```

jboddey (approved):

PR #921, opened 2024-10-25
Fix  security.services.vnc test

MariusBaldovin:
Increased maximum tcp port from 1000 to 10000
Added port 6001 (opened with vnc by x11 service ) to security.services.vnc test


modules/test/services/conf/module_config.json
jhughesbiot, line 1: Tested this change on a device and compared previous results and it took an additional 3 minutes with the longer scan.  Since we can't know for certain how long this new scan will take for every device, I think we should bump the default timeout up to at least 15 minutes (900).

MariusBaldovin: Updated the services module timeout to 900

Issue #920, opened 2024-10-25
Method to disable individual tests


duncangreene:
Raising issue further to e-mail conversation with @jboddey.

**Use case**
User should be able to disable individual tests from running, either from test pack (preferred), or from `module_config.json` files.

I tried two methods as instructed.

1. Deleting the highlighted block below from a test pack...

![Image](https://github.com/user-attachments/assets/c377aabf-5da3-4a6d-ba4b-1f71fb297a75)

...in this case the tests were still conducted, despite not being present in the test pack.

![Image](https://github.com/user-attachments/assets/8ae585e1-63d1-4163-acc2-758ddda35354)

2. Adding `"enabled": false` to the `module_config.json` file...

![Image](https://github.com/user-attachments/assets/6a58e9c6-d10c-42e0-88af-ad340fbdd559)

...yields the below in the report. I would maybe have expected a result of "Disabled" or for the tests to not appear at all, as opposed to "Feature Not Detected" (the _detection_ didn't happen).

![Image](https://github.com/user-attachments/assets/d8218715-3d1f-4f70-aa95-3f44e089f3e3)

These may warrant being two separate issues, but were discussed in the same thread of a need to disable individual tests. I can create a separate issue if wanted.

On a related point, it is noted that entire modules of tests can be disabled via the UI below, and so I'm unsure how the proposed disabling of individual tests may interact/interfere with the below in the UI. As an observation I would have thought test pack choice would override and even avoid the need to specify these options in the UI to some extent.

![Image](https://github.com/user-attachments/assets/c2f5efce-e0b2-4217-8c52-207955ce723e)

jhughesbiot:
When you change a module configuration file, you need to rebuild the modules for it to take effect. 

`sudo cmd/build`

This is not the recommended approach as it will disable these tests for all devices for all future tests and will also get overridden during upgrades.

The recommended approach is to modify the specific device configuration file.

For your example, if you navigate to the local/devices/< device folder >/device_config.json file and modify:

```
  "test_modules": {
    "protocol": {
      "enabled": true,
      "tests": {
        "protocol.valid_bacnet": {
          "enabled": false
        },
        "protocol.bacnet.version": {
          "enabled": false
        },
        "protocol.valid_modbus": {
          "enabled": false
        }
      }
    },
```

Which will give the following results:
![Image](https://github.com/user-attachments/assets/7009e41b-3073-4b67-865d-76bd948af7b2)


[...]

PR #956, opened 2024-11-07
Only run tests if they are included in the device's test pack

jboddey:
This PR prevents tests from running if they are not present in the chosen test pack.


jboddey:
@jhughesbiot We could do with passing the device config file, and test pack file to test modules - or we need the ability to control which tests a module runs. Any ideas?

MariusBaldovin (approved):

duncangreene:
Hey folks, out of interest how will the below be handled?

https://github.com/google/testrun/issues/920#issuecomment-2461769510

PR #960, opened 2024-11-11
Error handling for connection module

MariusBaldovin:
Added error handling if GRPC server can't start due to port 5001 being blocked by UFW firewall.

BEFORE

Module.log
Error_dhcp_before

Report
report_dhcp_ip_before

AFTER

Module.log
dhcp_error_handling_after

Report
1conn_before
1conn_before1

### DIFF
--- a/modules/test/conn/python/src/connection_module.py
+++ b/modules/test/conn/python/src/connection_module.py
@@ -464,7 +464,7 @@ class ConnectionModule(TestModule):
       LOGGER.error('Unable to connect to gRPC server')
       result = 'Error'
       description = (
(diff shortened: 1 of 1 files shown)


jhughesbiot:
I think the check firewall comments should be moved into the recommendations section of the test config rather than report it in the results.

jboddey:
> I think the check firewall comments should be moved into the recommendations section of the test config rather than report it in the results.

@jhughesbiot I agree, though I think we should cover that in a another PR as a separate feature (enabling steps to resolve for error results).

jhughesbiot (approved):

PR #966, opened 2024-11-13
Feature/UI/material3 theme

sofyakurilova:


jboddey:
Some findings from the branch so far:
 - Settings page: If the selected port is missing (resulting in two callouts at the top), then the save button on the settings page cannot be reached at 100% page zoom.
 - Certificates settings: The page is cut off at the bottom when in 100% zoom.
 - Device modal. Cannot edit a device. Always get the error: 'A device with that MAC address already exists'

Other non-blocking issues:
 - The device edit modal is very blue. No spacing between testing journey type and the icon.

OlgaMardvilko (approved):

jboddey:
When performing a test attempt, the device name appears to be a link (causing the cursor to change).

sofyakurilova:
@jboddey we want to fix the "welcome modal" in the next release http://b/413361299

jboddey (approved):

PR #979, opened 2024-11-20
Fix chrony install error

hitnik:


KrystsinaT (commented):
Verified. The package has been installed successfully.

PR #984, opened 2024-11-21
added --no-cache to for a fresh build

MariusBaldovin:
Added --no-cache to build process


jboddey:
Make this an optional argument when installing? Only to be called by cmd/install

jboddey (changes requested):
Can you add an optional argument in the cmd/build script that will add no-cache flag. The cmd/install script can then specify the optional arg when calling cmd/build so that this only affects user upgrades.

PR #1016, opened 2024-12-13
Jinja ntp report

hitnik:
report examples:
[report.pdf](https://github.com/user-attachments/files/18129984/report.pdf)

(rename pdf to html after download)
[report.html.pdf](https://github.com/user-attachments/files/18129992/report.html.pdf)


framework/python/src/core/session.py
```diff
@@ -658,7 +667,7 @@ def _remove_invalid_questions(self, questions):
         valid_questions.append(question)
 
       else:
-        LOGGER.debug(f'Removed unrecognised question: {question["question"]}')
+        LOGGER.debug(f'Removed unrecognised question: {question["question"]}') # pylint: disable=W1405
```
jboddey, line 670: Is the pylint disable necessary? We were never notified of a pylint issue before.

hitnik: I saw this error when running pylint.
Screenshot from 2024-12-16 12-44-37

OlgaMardvilko (approved):
The styles looks good to me, exactly the same as before.

framework/python/src/common/testreport.py
```diff
@@ -391,7 +410,7 @@ def _get_module_pages(self):
         if el.name == 'h1':
           current_size += 40 + h1_padding
         # Calculating the height of paired tables
-        elif (el.name == 'div'
+        elif (el.name == 'div' and el.has_attr('style')
```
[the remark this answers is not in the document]
hitnik: > Can you add an ID to the element and not reference it by style attr?

Done

PR #1017, opened 2024-12-16
Jinja service report

hitnik:
report examples:
[report.pdf](https://github.com/user-attachments/files/18151904/report.pdf)

(rename pdf to html after download)
[report.html.pdf](https://github.com/user-attachments/files/18151909/report.html.pdf)


OlgaMardvilko (approved):
LGTM

PR #1044, opened 2025-01-03
Added debug message in case users enters an incorrect device mac address

MariusBaldovin:
If users enter the incorrect device mac address, debug message will be displayed on terminal during the test with device mac address found but ignored and the expected device mac address.

### DIFF
--- a/framework/python/src/core/testrun.py
+++ b/framework/python/src/core/testrun.py
@@ -463,6 +463,8 @@ class Testrun:  # pylint: disable=too-few-public-methods
 
     if device is not None:
       if mac_addr != device.mac_addr:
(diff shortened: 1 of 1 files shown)


jboddey (approved):

framework/python/src/core/testrun.py
```diff
@@ -463,6 +463,8 @@ def _device_discovered(self, mac_addr):
 
     if device is not None:
       if mac_addr != device.mac_addr:
+        LOGGER.debug(f'Found device with mac addr: {mac_addr} but was ignored')
```
jboddey, line 466: Can you change this to info. Also, do we get any false MAC addresses appearing? Such as network containers or interfaces if not configured correctly?

PR #1050, opened 2025-01-07
Re-render device profile

MariusBaldovin:
When zip file is exported, if the device profile has been changed the report will be updated with new device profile data

### DIFF
--- a/framework/python/src/common/testreport.py
+++ b/framework/python/src/common/testreport.py
@@ -68,6 +68,9 @@ class TestReport():
     self._report_url = ''
     self._cur_page = 0
 
+  def update_device_profile(self, additional_info):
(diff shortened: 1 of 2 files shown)


jboddey:
@MariusBaldovin We also need to update the risk profile included in the ZIP file, I don't believe this currently happens?

MariusBaldovin:
> @MariusBaldovin We also need to update the risk profile included in the ZIP file, I don't believe this currently happens?

@jboddey I believed that risk profile is being updated since this addition is optional to the zip file. I will update the risk profile aswell.

MariusBaldovin:
@jboddey After testing, I can confirm that the risk profile included in the archive is also being updated when zip file is exported.

jboddey (approved):

PR #1052, opened 2025-01-08
Profile export endpoint

MariusBaldovin:
Created the /profile/{profile_name} enpoint. This returns the profile risk assessment in PDF format. If device is added into the body then the PDF contains device risk assessment. 

The PDF file is saved into 'local/risk_profiles' 

PDF with device:
risk assessment_device

PDF without device:
no_device_risk

### DIFF
--- a/framework/python/src/api/api.py
+++ b/framework/python/src/api/api.py
@@ -134,7 +134,9 @@ class Api:
     self._router.add_api_route("/profiles",
                                self.delete_profile,
                                methods=["DELETE"])
-    self._router.add_api_route("/profile/{profile_name}", self.get_profile)
+    self._router.add_api_route("/profile/{profile_name}",
+                               self.export_profile,
+                               methods=["POST"])
 
     # Allow all origins to access the API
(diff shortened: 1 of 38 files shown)


framework/python/src/api/api.py
```diff
@@ -926,6 +928,28 @@ async def delete_profile(self, request: Request, response: Response):
...
+    profile = self._session.get_profile(profile_name)
+
+    # If the profile not found return 404
+    if profile is None:
+      LOGGER.info("Profile not found, returning 404")
+      response.status_code = 404
+      return self._generate_msg(False, "Profile could not be found")
+
+    # Profile file path
+    profile_path = os.path.join(PROFILES_PATH, f"{profile_name}.json")
```
jboddey, line 942: We need to return the PDF, not the JSON

MariusBaldovin: It was added only for testing purposes, I am working on it

jboddey (approved):

Issue #904, opened 2024-10-18
Consider adding additional context to TLS server test names


duncangreene:
**What is the problem your feature is trying to solve?**
Multiple services on a device can run using differing versions of TLS. For example on a single device, the version and ciphers used by MQTT/TLS may be completely different from the version and ciphers used by HTTP/TLS.

The below shows a device that offers TLS v1.2 and TLS v1.3 on port 8009, but only TLS v1.2 on port 8443.

![Image](https://github.com/user-attachments/assets/c91a22f1-245e-48ba-a415-f50a3adbea88)

By the looks of the code I believe the `security.tls.v*_*_server` tests are specifically targeted at the service listening on port 443 on the DUT, and I wonder whether there would be merit in tightening the test name to reflect this? It may avoid the user incorrectly assuming that *all* services on the DUT offer 'TLS v1.3' when they read the result in the report for `security.tls.v1_3_server` (which as we can see above for 8443, isn't the case).

![Image](https://github.com/user-attachments/assets/cc8d5ab2-2813-4de5-a4d5-f966254223f4)

**Describe the solution you think would solve the problem**
Consider renaming to `security.tcp443.tls.v*_*_server` or similar.

**Additional context**
This would provide extensibility in the test namespace for additional TLS server tests to be carried out and individually reported against other ports.

- `security.tcp443.tls.v1_2_server`
- `security.tcp443.tls.v1_3_server`
- `security.tcp8883.tls.v1_3_server`
- `security.tcp8883.tls.v1_3_server`

jboddey:
I agree that we need to consider varying TLS implementations for different services - I am not sure the proposed test names are the right way of going about it - I will think further on another solution.

jboddey:
I think the solution for this would be a generic tls test which would result non-compliant if there were any TLS service that was considered non-compliant. The module report would then give you the information we want.

jboddey:
The TLS tests are being modified to specify the port numbers of invalid tls services. The test names will remain the same.

duncangreene:
@jboddey @jhughesoti Just on this subject, in 2.1.1 at least, am I correct in saying that the `security.tls.v*_*_server` tests are _only_ testing/reporting against ports found to offer HTTP/TLS?

That is, other ports offering services over TLS such as MQTT/TLS are not tested/reported on by the `security.tls.v*_*_server` tests?

jboddey:
Correct, just HTTP/TLS.

KrystsinaT:
[...]

PR #1055, opened 2025-01-10
Add full port scan to determine any HTTPS/TLS server running on non-s…

jhughesbiot:
Add a full port scan when determining if there are any web servers running and validate accordingly.

 - Check for both HTTP and HTTPS services running on all ports
 - Determines which services are HTTP and HTTPS
 - Fails TLS server tests if any HTTP ports are detected
 - Update report to indicate which ports have valid and invalid servers running
 - Increase module timeout by 2 minutes to account for port scan

Current report:
Screenshot 2025-01-10 130302

Updated report with HTTP port detected:
Screenshot 2025-01-10 140104

Updated report with valid web server:
Screenshot 2025-01-10 152655

Updated report with invalid web server:
Screenshot 2025-01-10 153104

### DIFF
--- a/modules/test/tls/conf/module_config.json
+++ b/modules/test/tls/conf/module_config.json
@@ -9,7 +9,7 @@
     "docker": {
       "depends_on": "base",
       "enable_container": true,
-      "timeout": 420
+      "timeout": 540
     },
     "tests":[
       {
--- a/modules/test/tls/python/requirements.txt
+++ b/modules/test/tls/python/requirements.txt
@@ -18,4 +18,4 @@ pyOpenSSL==24.3.0
 lxml==5.1.0 # Requirement of pyshark but if upgraded automatically above 5.1 will cause a
 pyshark==0.6
 requests==2.32.3
-
+python-nmap==0.7.1
--- a/modules/test/tls/python/src/http_scan.py
+++ b/modules/test/tls/python/src/http_scan.py
@@ -1,0 +1,76 @@
+# Copyright 2023 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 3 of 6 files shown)


MariusBaldovin:
The tls_report.html is not currently created. The issue is that tls.pcap if very large and contains over 134.000 packets. Processing the pcap file takes too long.
tls_killing_module

The solution is to filter by tls when tls.pcap is opened. 
filter_tls

jhughesbiot:
@MariusBaldovin Looks like that update reduced the time by about a minute and is generating fine for the device I was testing.  Does this work for you or do we need to increase timeout and/or work optimizing these tests further?

duncangreene:
From the brief look I had at https://github.com/google/testrun/pull/1055/commits/85fe62f4bc093290dc5ff3be9d98c1bbdfe7c77d, does this mean that the `security.tls.v*_*_server` tests will now specifically relate to HTTP/TLS only? And not other TLS-based services such as MQTT/TLS?

If so, I feel the test name perhaps needs tightening up to specifically pertain to HTTP/TLS? The subject of test naming and scope for the various TLS-based communication protocols is also discussed in #904.

MariusBaldovin:
> @MariusBaldovin Looks like that update reduced the time by about a minute and is generating fine for the device I was testing. Does this work for you or do we need to increase timeout and/or work optimizing these tests further?

It is fine the timeout set to 9 minutes. The TLS test took around 5 minutes.

jboddey:
> From the brief look I had at [85fe62f](https://github.com/google/testrun/commit/85fe62f4bc093290dc5ff3be9d98c1bbdfe7c77d), does this mean that the `security.tls.v*_*_server` tests will now specifically relate to HTTP/TLS only? And not other TLS-based services such as MQTT/TLS?
> 
> If so, I feel the test name perhaps needs tightening up to specifically pertain to HTTP/TLS? The subject of test naming and scope for the various TLS-based communication protocols is also discussed in #904.

Agreed. Our use case specifically is for http based services. I will start a discussion around non-http TLS services and see whether we can amend this PR to support the use case.

jboddey:
We will not look at non-http services in this PR, we may look at this in the future.

jboddey (approved):

Issue #1047, opened 2025-01-06
What tests or functionality are affected if running in --single-intf mode?


duncangreene:
What tests or functionality are affected if running in --single-intf mode?

https://github.com/google/testrun/blob/main/docs/get_started.md#:~:text=To%20run%20Testrun%20with%20just%20one%20interface%20(connected%20to%20the%20device)%2C%20use%20the%20%2D%2Dsingle%2Dintf%20option.

duncangreene:
Hey @jboddey @jhughesoti, I'm still not 100% clear on precisely which tests are affected (and how) by running in --single-intf mode.

The new disclaimer from #1056 will state the below.

>Tests that require an internet connection (e.g TLS client, DNS) will produce a non-compliant result.

**TLS client**
I imagine that DUT-to-Internet TLS connections will presumably go Feature Not Detected (as opposed to Non-Compliant) because the TCP 3 way handshake won't even complete?

Completely irrespective of --single-intf mode for a moment, I imagine the same applies though to any DUT that is configured to initiate TLS connections to non-Internet facing hosts (that obviously won't be present within the Testrun network)? That is, if a DUT is only configured to send syslog over TLS to a non-Internet facing syslog server, this (naturally) wouldn't be detected by Testrun anyway, and Testrun will register a `security.tls.v*_*_client` Feature Not Detected against the device in this case? Is this correct?

**DNS**
Given DNS is UDP, you'll presumably still see the DNS requests outbound even when running in --single-intf, surely these will still be picked up and reported by Testrun? The responses are somewhat superfluous in the context of detecting the DUT's ability to 'do' DNS?

![Image](https://github.com/user-attachments/assets/807e4a2e-36e9-4036-b339-97a45d0a439e)

Keen to hear thoughts, and also word on any other tests that you know to be implicated by running in --single-intf mode?

duncangreene:
@hitnik @sofyakurilova Please would you be able to advise on/answer the above?

duncangreene:
@hitnik @sofyakurilova Please would you be able to furnish an answer to the above original query? It is still not clear _exactly_ what tests or functionality are affected if running in --single-intf mode?

hitnik:
> [@hitnik](https://github.com/hitnik) [@sofyakurilova](https://github.com/sofyakurilova) Please would you be able to furnish an answer to the above original query? It is still not clear _exactly_ what tests or functionality are affected if running in --single-intf mode?

[...]

PR #1056, opened 2025-01-13
Add a disclaimer about tests affected by single interface mode

jboddey:


jboddey:
@duncangreene Does this change help with issue #1047?

duncangreene:
> @duncangreene Does this change help with issue #1047?

Hmmm... I mean, what tests would have an invalid result as a consequence of running in single interface mode? And would the results of said tests flag up as `invalid`? Or `skipped` or `disabled` from the below*?

*Unsure if this the complete list of valid statuses per se, but used for illustration.

image

MariusBaldovin (approved):

PR #1073, opened 2025-01-24
Generating module reports using Jinja2

hitnik:
report example 
[report.pdf](https://github.com/user-attachments/files/18535646/report.pdf)


jboddey:
In the example PDF provided, there are two separate TLS module module reports - one stating there are no certificates and the other showing certificate details. Also, the Outbound Connections title is indented which I don't believe it should be.

hitnik:
> Also, the Outbound Connections title is indented which I don't believe it should be.

The TLS module report was added by me specifically to show what a multi-page report module looks like. I fixed the Outbound Connections indent.

PR #1082, opened 2025-01-27
Fix protocol.services.bacnet 

MariusBaldovin:
Port 47808 was not scanned. Added protocol.services.bacnet  in test_packs. 

Before
scan_before

After
protocol services bacnet

### DIFF
--- a/resources/test_packs/qualification.json
+++ b/resources/test_packs/qualification.json
@@ -155,7 +155,7 @@
         },
         {
           "name": "protocol.services.bacnet",
-          "required_result": "Required"
+          "required_result": "Informational"
         },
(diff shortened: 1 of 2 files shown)


resources/test_packs/qualification.json
```diff
@@ -153,6 +153,10 @@
           "name": "ntp.network.ntp_server",
           "required_result": "Required"
         },
+        {
+          "name": "protocol.services.bacnet",
+          "required_result": "Required"
```
jboddey, line 158: This should be informational for Qualification too.

jboddey (approved):

Issue #1087, opened 2025-01-28
Services Module shows no open ports when ports are open


duncangreene:
This might be regarded as the same issue as #1080, but raising as a separate issue just in case the Services vs. TLS output covered in #1080 might be regarded as a separate issue.

**Describe the bug**
Services module shows no open ports...

![Image](https://github.com/user-attachments/assets/4ca4bb9d-10f2-48fc-b3e4-186f30fd4b13)

...but a standalone nmap scan against the same device yields the below open ports.

```
PS C:\Users\Lenovo ThinkPad> nmap -sTUV --top-ports 1000 10.225.10.179
Starting Nmap 7.93 ( https://nmap.org/ ) at 2025-01-28 11:40 GMT Standard Time
NSOCK ERROR [0.0580s] ssl_init_helper(): OpenSSL legacy provider failed to load.
 
Warning: 10.225.10.179 giving up on port because retransmission cap hit (10).
Nmap scan report for 10.225.10.179
Host is up (0.0030s latency).
Not shown: 994 filtered tcp ports (no-response), 908 closed udp ports (port-unreach), 87 open|filtered udp ports (no-response)
PORT      STATE SERVICE        VERSION
22/tcp    open  ssh            OpenSSH 8.2 (protocol 2.0)
23/tcp    open  telnet         BusyBox telnetd
53/tcp    open  domain         dnsmasq 2.86
80/tcp    open  http           nginx
443/tcp   open  ssl/http       nginx
8080/tcp  open  ssl/http-proxy
53/udp    open  domain         dnsmasq 2.86
123/udp   open  ntp            NTP v4 (secondary server)
161/udp   open  snmp           SNMPv1 server; net-snmp SNMPv3 server (public)
1701/udp  open  L2TP?
47808/udp open  bacnet
1 service unrecognized despite returning data. If you know the service/version, please submit the following fingerprint at https://nmap.org/cgi-bin/submit.cgi?new-service :
SF-Port8080-TCP:V=7.93%T=SSL%I=7%D=1/28%Time=6798CBDC%P=i686-pc-windows-wi
SF:ndows%r(GenericLines,67,"HTTP/1\.1\x20400\x20Bad\x20Request\r\nContent-
SF:Type:\x20text/plain;\x20charset=utf-8\r\nConnection:\x20close\r\n\r\n40
SF:0\x20Bad\x20Request")%r(GetRequest,2D9,"HTTP/1\.0\x20404\x20Not\x20Foun
[...]

Issue #1080, opened 2025-01-27
Services Module shows no open ports but TLS Module shows cert on port 443


duncangreene:
**Describe the bug**
Services Module shows no open ports...

![Image](https://github.com/user-attachments/assets/f369e19b-2e21-46c3-905f-4102374f6c88)

...but TLS Module shows information for cert presented on port 443.

![Image](https://github.com/user-attachments/assets/5199dbce-bf9b-49f0-b89e-b71f7fcf9d74)

**To Reproduce**
Steps to reproduce the behaviour:
1. Run test against device running HTTPS on port 443. DUT in this case is [this](https://www.milesight.com/iot/product/lorawan-gateway/ug67).

**Expected behaviour**
Services Module would show port 443 open.

**Error logs**
[test_result.zip](https://github.com/user-attachments/files/18557898/milesightug67.zip)

**Environment (please provide the following information about your setup):**
 - Version: 2.1

**Additional context**
Only modifications from official 2.1 release are as below.

- resources/test_packs/qualification.json (Set tests to excluded/recommended/informational as appropriate)
- resources/devices/device_profile.json (Add custom questions)

jhughesbiot:
On review of the logs provided, it appears the nmap scan is not detecting any ports open where the TLS tests directly connect to port 443 without any scan.  Is it possible the device has some sort of service or firewall that blocks port scanning?

duncangreene:
There's no firewall that blocks port scanning per se, and a port scan against this device is available in the potentially related issue #1087.

PR #1096, opened 2025-01-31
Fix services resolving wrong ip

jhughesbiot:
Some devices resolve an IP after connection module and then quickly change after nmap has started causing wrong port scanning results.  This forces an ip check before nmap and that the services module runs before connection module so the IP should be more stable before DHCP behavior is tested.

Bugs reporting this issue:
https://github.com/google/testrun/issues/1080
https://github.com/google/testrun/issues/1087


jboddey (approved):

jboddey:
I agree that we need to resolve the IP address before the scan. I'm not sure if re-ordering the modules is necessary. Can the re-ordering part be removed? And then can we add a new key-value in the module-config called priority, so that we can manage loading order there instead of hardcoding module names?

jhughesbiot:
> Can the re-ordering part be removed?

Unfortunately no, I have tested with and without this and if it comes after the conn module, some devices keep consistent IP's and some don't.  The only stable way was to enforce the order.

> And then can we add a new key-value in the module-config called priority, so that we can manage loading order there instead of hardcoding module names?

This is likely the best solution but we should do this in a separate PR since this would also impact the protocol module.

jboddey (approved):

PR #1168, opened 2025-03-03
Refactor FE to use BE endpoint to get the links for the report

sofyakurilova:
Use field "export" for zip download, or make url from "report" field if "export" is missing

Video after changes. Please note that mocks are used, so the link is not actually working
http://screencast/cast/NDg0Mzc5OTYxOTU2NzYxNnw5Yzk5NWI3OC1lOA


OlgaMardvilko (changes requested):
Should we make the export field to have the dynamic API URL (the same as report field)?
The code is here: https://github.com/google/testrun/blob/dev/modules/ui/src/app/services/test-run.service.ts#L60

PR #1243, opened 2025-04-16
Removed Ubuntu 20.04 runner

MariusBaldovin:
Ubuntu 20.04 runner is not supported anymore on Github.


jboddey:
Could you update the docs to state we do not support Ubuntu 20 anymore please?

jboddey (approved):

PR #1244, opened 2025-04-16
410755779: (fix) changes to move focus to the first interactive elem on opened page

OlgaMardvilko:
### Overview

Ticket: 410755779
fix: changes to move focus to the first interactive elem on opened page

### Video after changes
https://screencast.googleplex.com/cast/NjI3NjQwMjEzNDM4NDY0MHxjY2NkMDFhNy02YQ

### DIFF
--- a/modules/ui/src/app/app.component.ts
+++ b/modules/ui/src/app/app.component.ts
@@ -326,7 +326,8 @@ export class AppComponent implements AfterViewInit {
         take(1)
       )
       .subscribe(() => {
-        this.appStore.setFocusOnPage(null);
+        const mainContainer = window.document.querySelector('#main');
+        this.appStore.setFocusOnPage(mainContainer);
       });
   }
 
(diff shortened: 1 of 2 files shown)


modules/ui/src/app/app.component.ts
```diff
@@ -326,7 +326,8 @@ export class AppComponent implements AfterViewInit {
         take(1)
       )
       .subscribe(() => {
-        this.appStore.setFocusOnPage(null);
+        const mainContainer = window.document.querySelector('#main');
+        this.appStore.setFocusOnPage(mainContainer);
```
sofyakurilova, line 330: Maybe we don't need to pass any argument here?  
In focus-manager.service the container will be #main in this case
  focusFirstElementInContainer(
    container: Document | Element | null = window.document.querySelector(
      '#main'
    )
  )

OlgaMardvilko: > Maybe we don't need to pass any argument here? In focus-manager.service the container will be #main in this case focusFirstElementInContainer( container: Document | Element | null = window.document.querySelector( '#main' ) )

Right, I saw it, but it doesn't work.

sofyakurilova (approved):

Issue #1228, opened 2025-04-09
security.tls.v1_2_server reporting certificate invalid even though DUT does not support TLS 1.2


duncangreene:
**Describe the bug**
security.tls.v1_2_server is reporting 'certificate invalid'...

![Image](https://github.com/user-attachments/assets/f3ed318a-6180-4c24-9f90-c86028f6ae1c)

...even though the DUT does not support TLS 1.2...

![Image](https://github.com/user-attachments/assets/b19b951c-c742-4cee-a6ab-4a6ebc71d1be)

**Expected behavior**
Words to the effect of 'TLS 1.2 not supported' would be shown.

**Error logs**
[report (33).zip](https://github.com/user-attachments/files/19665057/report.33.zip)

**Environment (please provide the following information about your setup):**
 - OS: Ubuntu 22.04.5
 - Version: 2.1.1 (modifications from official 2.1.1 release are as below)
    - make/DEBIAN/control (Mod version number to append "-oss.1.0")
    - resources/test_packs/qualification/config.json (Set tests to excluded/recommended/informational as appropriate)
    - resources/devices/device_profile.json (Add custom questions)

**Additional context**
For info, that nmap script **will** report other TLS versions (on any port) if they are supported. Example below.

![Image](https://github.com/user-attachments/assets/f0a5b573-c7e9-4788-813e-4e25f55a9f9f)

PR #1248, opened 2025-04-16
Fix bugs in _security_tls_v1_2_server

MariusBaldovin:
Fix the bugs in _security_tls_v1_2_server
Added unit tests for _security_tls_v1_2_server


modules/test/tls/python/src/tls_module.py
```diff
@@ -344,8 +344,12 @@ def _security_tls_v1_2_server(self):
...
             tls_1_3_results = self._tls_util.validate_tls_server(
                 host=self._device_ipv4_addr, port=port, tls_version='1.3')
-            port_results = self._tls_util.process_tls_server_results(
-                tls_1_2_results, tls_1_3_results, port=port)
+            # If TLS 1.2 is not supported don't process the results
+            if tls_1_2_results[0] is not None:
+              port_results = self._tls_util.process_tls_server_results(
+                  tls_1_2_results, tls_1_3_results, port=port)
+            else:
+              port_results = None
```
jhughesoti, line 352: TLS 1.3 can still validate TLS 1.2.  If a device is running just TLS 1.3, it shouldn't fail TLS 1.2 since the requirement is at least TLS 1.2 for this test so there should be no need for this additional check.

MariusBaldovin: _tls_util.process_tls_server_results is causing a bug when TLS 1.2 is not supported the TLS 1.3 will incorrectly validate 1.2 (https://github.com/google/testrun/issues/1228). With these changes when TLS 1.2 is not supported the test result will be  "Feature not detected" and since this test is "Required if Applicable" it will not affect the overall test result. The TLS 1.3 will still validate TLS 1.2 but only when 1.2 is non compliant and 1.3 is compliant (line 374). I have added unit tests for all use cases for  _security_tls_v1_2_server.

jhughesoti: That bug is only if TLS 1.3 is invalid and TLS 1.2 is valid so this still needs to check both TLS 1.2 and 1.3 in the 1.2 test but not fail if TLS 1.2 is valid but not 1.3.  This needs to be fixed in the validation logic, not removing that scenario from being tested.

jboddey:
tls_deprecated
If TLS v1.1 or below is present: Non-Compliant

tls_ciphers
TLS v1.2, if present, must use the correct ciphers
 
tls_v1_3
TLS v1.3, informational, is it present?

tls_min_version (new test): Compliant if v1.2 or v1.3 present

jhughesoti (approved):

PR #1295, opened 2025-05-13
Testing/modbus

MariusBaldovin:
Added Compliant result for protocol.valid_modbus in test_tests


jboddey:
@MariusBaldovin could the naming of files be changed to protocol rather than bacnet now that it supports both?

jboddey (approved):

PR #1394, opened 2025-12-22
The Non-complient test with Required if Applicable value is not highlighted in the results list on UI 

sofyakurilova:
Highlight non-compliant test result; update version (the current release)


hitnik:
[report.html.pdf](https://github.com/user-attachments/files/24310992/report.html.pdf)
[report.pdf](https://github.com/user-attachments/files/24310993/report.pdf)

PR #1419, opened 2026-02-03
Test Module Report - TLS Module

sofyakurilova:
Generate html report for tls module


hitnik:
[report (17).zip](https://github.com/user-attachments/files/25099662/report.17.zip)

PR #1506, opened 2026-04-29
The message about unexpected device connected doesn' t appear

sofyakurilova:
Join log messages for mqtt


KrystsinaT (approved):
Thank you!

Issue #1290, opened 2025-05-07
Sporadic results in Services tests


duncangreene:
**Describe the bug**
In 2.2 I'm still seeing some sporadic results in Services tests. This issue was initially uncovered in #885 and seems to occur on more resource constrained devices, which are prevalent (and constrained _by design_) in the IoT/smart building world.

![Image](https://github.com/user-attachments/assets/2983f80b-a514-4556-9775-85dd2e23fea5)

**Test A**
![Image](https://github.com/user-attachments/assets/2f035405-6b99-4364-b7b3-5d4a5fb50541)

**Test B**
![Image](https://github.com/user-attachments/assets/1da8f66d-436c-4ce4-9878-9c2af5cbf2d3)

**Test C**
![Image](https://github.com/user-attachments/assets/64466248-f246-49e2-998e-e17303696c0a)

**Test D**
![Image](https://github.com/user-attachments/assets/ca224607-f02d-4a79-8b6a-064027e37190)

**Expected behavior**
The same device under the same conditions should yield the same results every time.

**Error logs**
[testA.zip](https://github.com/user-attachments/files/20087767/testA.zip)
[testB.zip](https://github.com/user-attachments/files/20087768/testB.zip)
[testC.zip](https://github.com/user-attachments/files/20087770/testC.zip)
[testD.zip](https://github.com/user-attachments/files/20087776/testD.zip)

**Environment (please provide the following information about your setup):**
 - Version 2.2

**Additional context**
There was discussion on #885 previously about utilising nmap's inbuilt rate limiting features to mitigate the false negatives/false positives that can/will occur against devices currently.

The comment [here ](https://github.com/google/testrun/issues/885#issuecomment-2587628036)states that a second HTTP specific scan is carried out in 2.2 that may well resolve false negatives/false positives with the HTTP related tests, but this naturally doesn't cover all the other possible service/port permutations out there (TCP 502 in the case of the device above, but that's just one device/service in this example).

duncangreene:
My two cents on the mention of 'constrained by design' devices above.

[...]

Issue #885, opened 2024-10-15
security.services.http Sporadic Missing HTTP Requests


duncangreene:
**Describe the bug**
Sporadically (haven't been able to deduce how often), some HTTP Requests (I think just HTTP 1.1) are seen missing from the services.pcap.

![Image](https://github.com/user-attachments/assets/767e89d9-3b95-4db5-adea-0dfef5ab326a)

This causes a false "No HTTP server found" to be recorded in the report, despite the DUT running a HTTP server.

![Image](https://github.com/user-attachments/assets/884a489e-1293-46ca-9407-64ab84421d9b)

**To Reproduce**
1. Run services test.

**Expected behavior**
All HTTP Requests should be present.

![Image](https://github.com/user-attachments/assets/41fe127f-f743-4c2e-b96c-b0f60e8c057c)

**Error logs**
[allhttprequests.zip](https://github.com/user-attachments/files/17377469/allhttprequests.zip)
[missinghttprequests.zip](https://github.com/user-attachments/files/17377468/missinghttprequests.zip)

**Environment (please provide the following information about your setup):**
 - OS: Ubuntu
 - Version 24.04.1
 - Testrun 2.0

**Additional context**
DUT is https://www.ismacontrolli.com/en/isma-b-mix18-ip.html

jhughesbiot:
The presence of HTTP traffic does not indicate an HTTP server running on the device.  For that first example you provided, the traffic detected is to port 502, which is modbus traffic.  I confirmed modbus traffic is present on the report in the protocol tests. 

In the second example, confirming with the report, it does detect the HTTP server being open.  This could be a timing issue with the device in question and the testing steps.  i.e., example one tested services module before the device was in a steady state where the web server was ready vs the second where it was. 


Any additional steps for testing to produce this would be helpful in determining the root cause.

duncangreene:
Aha, I hadn't registered the port number in the first 3 telegrams.

The issue appears to be the absence of the outgoing requests from Testrun (10.10.10.9) to port 80 in the first example. What would cause Testrun to not send these outgoing requests?

duncangreene:
The original post contains the two test .zip outputs, if the log files you're after are included in there? Let me know if there are other logs etc. to include.

I can see from the services.pcap in missinghttprequests.zip that the DUT doesn't SYN ACK to the port 80 SYN, however this device does indeed run a bona fide web server as part of its 'normal' operation (and as evidenced in the respective services.pcap in allhttprequests.zip).

[...]

PR #1521, opened 2026-05-21
Use gentler nmap timing

duncangreene:
To address #1290 and #885. As demonstrated in both issues, Testrun has been shown to routinely exhibit false negatives for devices by scanning too aggressively. The timing modifications in the attached were tuned against a constrained device that was routinely being flagged with false negatives.

Using the attached on my various Testrun machines increases TCP scan time from ~30 seconds (with false positives) to ~3 minutes (with no false positives) in isolation, and UDP scan time remains unchanged in isolation because of the small amount of UDP ports being scanned (the bulk of the time is spent carrying out the version detection step, as opposed to the port range scan itself).

False negatives can still occur if this PR is not combined with #1520, which forces the TCP and UDP scans to happen consecutively (minimal risk of false negatives), instead of concurrently (increased risk of false negatives).

### DIFF
--- a/modules/test/services/python/src/services_module.py
+++ b/modules/test/services/python/src/services_module.py
@@ -213,16 +213,15 @@ class ServicesModule(TestModule):
     # Scans TCP ports to detect open ports.
     host = self._device_ipv4_addr
     # Preliminary command for quick open port detection.
-    nmap_command_prelim = f'''nmap --open -sT -Pn -v -n -T5
-                          --version-intensity 0 --min-rate 1000
-                          -p- -oX - {host}'''
+    nmap_command_prelim = f'''nmap --open -sT -Pn -n -v -T3
+                          --scan-delay 2ms -p- -oX - {host}'''
     LOGGER.info(f'Running nmap TCP port scan for {host}')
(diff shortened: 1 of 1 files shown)


duncangreene:
I am currently investigating performance of this when it’s running in a VM under VirtualBox as I’m finding the preliminary TCP scan is taking an incredibly long time. I suspect it’s a confluence of the removal of min-rate and addition of scan-delay args, but will aim to fully diagnose.

Running this PR on bare metal is unaffected by the above on my different machines so far.

duncangreene:
> I am currently investigating performance of this when it’s running in a VM under VirtualBox as I’m finding the preliminary TCP scan is taking an incredibly long time. I suspect it’s a confluence of the removal of min-rate and addition of scan-delay args, but will aim to fully diagnose.
> 
> Running this PR on bare metal is unaffected by the above on my different machines so far.

Original PR still stands. It transpires the above was simply down to an under-resourced VM. Allocating more CPU and allowing VirtualBox to utilise VT-x/AMD-V resolved it, and gave similar performance to that of bare metal.

PR #1547, opened 2026-06-17
Improve Configuration Handling and Runtime Validation

shikha-sarraf:
Adding a configurable ui_port setting (default: 8080) for UI container startup.
Implementing a safe timezone fallback to UTC when host timezone detection is unavailable.
Validating Docker container PIDs before use to prevent invalid PID-related failures.
Replacing hardcoded private network values with configurable subnet and IP range parameters.
Improving logger configuration parsing with graceful fallback behavior.
Replacing a broad exception catch with specific OSError and ValueError handling.

### DIFF
--- a/framework/python/src/api/api.py
+++ b/framework/python/src/api/api.py
@@ -550,8 +550,7 @@ class Api:
       response.status_code = 200
       return self._generate_msg(True, "Successfully deleted the device")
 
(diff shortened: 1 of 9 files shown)


shikha-sarraf (commented):
Checked , all good ready for merge

PR #1552, opened 2026-06-24
Adopting a name according to Timestamp Closes #1511

### LINKED ISSUE #1511: Consider adopting a concatenation of Manufacturer Model Firmware and Timestamp as the report .zip filename
**What is the problem your feature is trying to solve?**
Currently when the .zip file is downloaded it is named "report.zip". As a result you naturally end up with a lot of the below for different tests.

Image

**Describe the solution you think would solve the problem**
Testrun should adopt a concatenation of the manufacturer, model, firmware and test timestamp as the filename, for easy identification 'at a glance'.

sofyakurilova:
Use the device data for *.zip file name


hitnik (approved):

duncangreene:
@sofyakurilova @hitnik did you see the comment about potentially including the MAC address on #1511 [here](https://github.com/google/testrun/issues/1511#issuecomment-4912816774)?

PR #1566, opened 2026-07-02
feat: add optional static IP support for devices without DHCP

### LINKED ISSUE #1450: device has static IP support only
I'm bringing two issues to your attention regarding the current test run.

First, the DUT does not support dynamic IP assignment and only allows for static IP assignment. Due to this limitation, the test run cannot obtain the DUT's IP address.

Additionally, I am experiencing a connectivity issue on the Linux PC. After running the Docker container, the machine is no longer able to connect to the internet.

IMINABO1:
Closes #1450

Some devices (e.g. industrial/elevator controllers) don't support DHCP and currently cause Testrun to cancel the run when no IP is obtained.

This PR adds optional static IP support: if a device's `device_config.json` includes an `"ip_addr"` field, the network orchestrator skips the DHCP wait and uses that IP directly. Devices without `ip_addr` are completely unaffected, existing behaviour is unchanged.

Changes:
- `device_config.json` accepts an optional `"ip_addr"` (validated as IPv4)
- Network orchestrator skips DHCP startup capture for static-IP devices
- `_dhcp_lease_ack` no longer overwrites a pre-configured static IP
- Unit tests for the new static IP path (`testing/unit/framework/static_ip_test.py`)
- CI action version bumps (Node 20 → 24)


IMINABO1:
I have signed the CLA.

IMINABO1 (commented):
Updated the PR with a fix for the report migration failure:

- Added export URL handling for migrated reports
- Preserved report/export metadata in serialized device config
- Added regression coverage

Tested with:
`python testing\unit\framework\static_ip_test.py -v`

duncangreene:
@IMINABO1 I've not studied the PR in depth, but does this require the user to manually add an IP address into the device_config.json following device addition within the Testrun UI?

Does it cater for _any_ private IP address? E.g. Would it work if the device had IP address 192.168.0.32/24 (i.e. outside of the default Testrun network range of 10.10.10.0/24)?

I have been working on a similar feature that doesn't require any manual file modifications, but it does predicate that the DUT must have a prescribed static IP address to start with. It's not quite ready for release yet, but has been working well in testing so far.

IMINABO1:
@duncangreene  The latest commit (97a6160) fixes the CI failures on this PR (unit test pollution and mixed line endings). The workflow runs are now waiting on maintainer approval. could you approve them when you get a chance? Thanks!

IMINABO1:
@duncangreene Good questions, answering both based on what's in the PR today:

**1. Manual edit of `device_config.json`?** Yes, currently. After adding the device through the UI, you add an `"ip_addr"` field to that device's `device_config.json`. It's validated as IPv4 at load time (invalid values are logged and ignored, falling back to the normal DHCP path) and preserved through config re-serialization. I deliberately kept this PR's scope to the orchestrator so devices without `ip_addr` are completely unaffected, but I'm happy to add a UI input for it as a follow-up if that's the direction you'd prefer.

**2. Any private IP?** Validation accepts any valid IPv4, and the orchestrator skips the DHCP wait and uses it as-is. Practically though, the device still needs to be reachable on the Testrun network, so an address within the default 10.10.10.0/24 subnet is what works today. 192.168.0.32/24 would pass validation but test modules wouldn't reach it without extra routing on the Testrun interface. I can document that constraint, or add subnet handling as a follow-up.

Since your feature predicates on the DUT having a prescribed static IP, the two approaches sound compatible. This PR covers the case where the device can't do DHCP at all, with minimal surface area. Happy to adapt it to fit the direction you're planning.

IMINABO1:
@duncangreene @sofyakurilova  Following up on this one.

The workflow runs from August sat waiting for approval and expired after 30 days, so the red checks aren't test failures. The Pylint and Unit failures from the earlier run were fixed in 97a6160.

I've also retargeted this PR to `dev`, since that's where the project takes PRs, and merged the latest `dev` into the branch with no conflicts. The full test suite runs on a schedule in my fork and passes on this commit.

Could you approve the workflow runs when you get a chance? Happy to make changes on the static IP approach if you have feedback on my answers above.

PR #1622, opened 2026-08-06
Fix TLS 1.3 text appearing in security.tls.v1_2_server results

### LINKED ISSUE #1529: Result description pertaining to TLS 1.3 appearing in security.tls.v1_2_server
**Describe the bug**
A result description pertaining to TLS v1.3 is appearing in the result description for security.tls.v1_2_server.

Image

**Expected behavior**
A result description pertaining to TLS v1.3 is appears in the result description for security.tls.v1_3_server.

**Error logs**
[report - 2026-06-01T105325.608.zip](https://github.com/user-attachments/files/28460270/report.-.2026-06-01T105325.608.zip)

**Environment (please provide the following information about your setup):**
 - Version 2.3.3

**Additional context**
This was previously observed too in this post [here](https://github.com/google/testrun/issues/1389#issuecomment-3603321891) on the same DUT from Dec 2025. Device in question is a Tridium JACE 8000.

IMINABO1:
TLS 1.3 detail lines were always appended to the TLS 1.2 server test output. Only include them when a valid TLS 1.3 certificate is what makes an invalid TLS 1.2 result compliant, so the result description for security.tls.v1_2_server otherwise only reports TLS 1.2.

Updated the unit tests in testing/unit/tls/tls_module_test.py that pinned the old output; the intentional 'TLS 1.2 invalid + TLS 1.3 valid: Compliant' case still reports both versions since the TLS 1.3 result explains the verdict.

Fixes #1529

### DIFF
--- a/modules/test/tls/python/src/tls_util.py
+++ b/modules/test/tls/python/src/tls_util.py
@@ -484,11 +484,14 @@ class TLSUtil():
         details.extend(tls_1_2_results[1])
       else:
         details.append(tls_1_2_results[1])
-      details.append(f'TLS 1.3 {positive_1_3}validated on port {port}:')
-      if isinstance(tls_1_3_results[1], list):
-        details.extend(tls_1_3_results[1])
-      else:
-        details.append(tls_1_3_results[1])
+      # Include TLS 1.3 details only when a valid TLS 1.3 certificate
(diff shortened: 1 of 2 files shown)


duncangreene:
I believe this may already be resolved by #1598.

PR #1640, opened 2026-09-10
The search chips display

sofyakurilova:
Adds Assessment type filter; fix Started filter display; the chip value is not truncated


KrystsinaT (approved):
Approved

